PR #176, opened 2020-03-18
support multiple time queryables

azahnen:


ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ldproxy/ogcapi/features/core/application/OgcApiFeaturesCoreConfiguration.java
```diff
@@ -34,6 +35,9 @@
     static final int MINIMUM_PAGE_SIZE = 1;
     static final int DEFAULT_PAGE_SIZE = 10;
     static final int MAX_PAGE_SIZE = 10000;
+    static final String PARAMETER_BBOX = "bbox";
+    static final String PARAMETER_DATETIME = "datetime";
+    static final String DATETIME_INTERVAL_SEPARATOR = "...";
```
cportele: It is just an internal convention, so the exact string is not important, but in general "/" would be more intuitive as the slash is the separator in time intervals in ISO 8601.

ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ldproxy/ogcapi/features/core/application/OgcApiFeaturesQueryImpl.java
```diff
@@ -1,6 +1,6 @@
 /**
  * Copyright 2020 interactive instruments GmbH
- *
+ * <p>
```
cportele: This shouldn't be here.

ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ldproxy/ogcapi/features/core/application/OgcApiFeaturesQueryImpl.java
```diff
@@ -223,21 +229,21 @@ public FeatureQuery requestToFeatureQuery(OgcApiApiDataV2 apiData, FeatureTypeCo
...
-                                                     Optional<EpsgCrs> providerCrs, Cql.Format cqlFormat) {
+                                                  Map<String, String> filterableFields, Set<String> filterParameters,
+                                                  Optional<EpsgCrs> providerCrs, Cql.Format cqlFormat) {
 
         List<CqlPredicate> predicates = filters.entrySet()
                                                .stream()
                                                .map(filter -> {
                                                    if (filter.getKey()
-                                                             .equals("bbox")) {
+                                                             .equals(OgcApiFeaturesCoreConfiguration.PARAMETER_BBOX)) {
```
cportele: The other constants are imported, but this one isn't?

Issue #173, opened 2020-03-16
review / improve exception handling


cportele:
API exceptions should be thrown in the top layer only. Exceptions thrown from other modules should be caught in the top level and handled consistently.

PR #196, opened 2020-06-29
doc: exceptions

cportele:
related to #173


azahnen:
It might be good to add a rule for exception logging. I think this should mostly happen in the web-layer classes as well. I use to log the message with level error (if it really is an error) and the stacktrace with level debug, like this:

```
LOGGER.error("Feature provider with id '{}' could not be started: {}", getId(), e.getMessage());
if (LOGGER.isDebugEnabled()) {
    LOGGER.debug("Stacktrace:", e);
}
```

Issue #163, opened 2020-03-15
cleanup non-ogcapi modules


azahnen:
Remove/move ldproxy-admin-rest, ldproxy-config, ldproxy-service-legacy-migration and ldproxy-wfs3.

PR #200, opened 2020-07-27
Add centralized exception handling

ulugbek:


ulugbek:
A few additional points regarding the rule 3 from the [documentation](https://github.com/interactive-instruments/ldproxy/blob/master/docs/dev/exceptions.md):
a) Is the rule also applied to OpenApi classes? For instance, ServerErrorException is thrown by OpenApiHtml.
b) CodelistEndpoint also throws BadRequest exceptions, but it's not a subclass of OgcApiEndpointExtension. Should we keep it this way or change the exception?
c) Should it be allowed to throw WebApplicationExceptions (specifically Bad Request) by the subclasses of OgcApiQueryParameter and OgcApiPathParameter?

cportele:
> a) Is the rule also applied to OpenApi classes? For instance, ServerErrorException is thrown by OpenApiHtml.
> c) Should it be allowed to throw WebApplicationExceptions (specifically Bad Request) by the subclasses of OgcApiQueryParameter and OgcApiPathParameter?

Good question whether we want to allow WebApplicationExceptions also from format classes (like OpenApiHtml) or parameter classes. In general, I would say that we should to keep it to the current list in rule 3. 

> b) CodelistEndpoint also throws BadRequest exceptions, but it's not a subclass of OgcApiEndpointExtension. Should we keep it this way or change the exception?

CodelistEndpoint is a special case and something to clean-up as part of #163. I think it is ok to keep it as it is for now.

cc @azahnen

ogcapi-stable/ogcapi-common/src/main/java/de/ii/ldproxy/ogcapi/infra/rest/OgcApiExceptionMapper.java
```diff
@@ -0,0 +1,124 @@
...
+    @Context
+    private UriInfo uriInfo;
+
+    @Validate
+    void onStart() {
+        LOGGER.debug("OGC API EXCEPTION WRITER");
+    }
+
+    @Override
+    public Response toResponse(Throwable exception) {
```
cportele, line 62: This looks as if everything that it not a WebApplicationException is mapped to a 500, which is not correct. Some of the exceptions need to be mapped to 400, 404, etc. Or am I missing something?

ogcapi-stable/ogcapi-common/src/main/java/de/ii/ldproxy/ogcapi/infra/rest/OgcApiExceptionMapper.java
```diff
@@ -0,0 +1,124 @@
...
+    @Override
+    public long getSize(OgcApiErrorMessage e, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return -1;
+    }
+
+    @Override
+    public void writeTo(OgcApiErrorMessage e, Class<?> type, Type genericType, Annotation[] annotations,
+                        MediaType mediaType, MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream)
+            throws IOException, WebApplicationException {
+        if (MediaType.TEXT_HTML.equals(mediaType.toString())) {
```
cportele, line 105: JSON and HTML output of exceptions should also be modelled like all other formats, as separate classes that implement FormatExtension.

ogcapi-draft/ogcapi-observation-processing/src/main/java/de/ii/ldproxy/ogcapi/observation_processing/parameters/QueryParameterCoordPosition.java
```diff
@@ -147,7 +146,7 @@ public boolean isEnabledForApi(OgcApiApiDataV2 apiData, String collectionId) {
         if (coord==null) {
             coord = getDefault(apiData, Optional.of(featureType.getId())).orElse(null);
             if (coord == null)
-                throw new BadRequestException(String.format("The required parameter '%s' has no value.", getName()));
+                throw new IllegalStateException(String.format("The required parameter '%s' has no value.", getName()));
```
cportele, line 149: If a parameter value is missing/incorrect, this sounds like an IllegalArgumentException to me. 

IllegalStateException is "signals that a method has been invoked at an illegal or inappropriate time", which is not the reason for this issue.

I think we probably should add [exception guidance](https://github.com/interactive-instruments/ldproxy/blob/master/docs/dev/exceptions.md) how to map situations to standard Java exceptions and how these are eventually mapped to Web exceptions. Could you maybe summarize how you mapped which situations as a starting point?

ogcapi-stable/ogcapi-common/src/main/java/de/ii/ldproxy/ogcapi/infra/rest/OgcApiExceptionMapper.java
ulugbek: Yes, you're right. So as you suggested below, I will summarize how to map different error scenarios to Java exceptions and how they in turn are mapped to Web exceptions

ogcapi-stable/ogcapi-common/src/main/java/de/ii/ldproxy/ogcapi/infra/rest/OgcApiExceptionMapper.java
ulugbek: Okay, will update this

ogcapi-stable/ogcapi-common/src/main/java/de/ii/ldproxy/ogcapi/infra/rest/OgcApiExceptionFormatJson.java
```diff
@@ -25,18 +23,17 @@
     @Requires
     SchemaGenerator schemaGenerator;
```
ulugbek: JSON responses fail because the SchemaGenerator is always null. Any ideas why this could be happening?

cportele: I think, because SchemaGenerator is not ready when the class is created. I have changed it to create the simple RFC 7807 schema manually.

PR #201, opened 2020-08-03
Store redesign

### LINKED ISSUE #158: refactor store/settings
The directory `store/settings` will not be used any more. 

Settings that concern services will be moved to the according buildingBlock in the service configuration (`geojson`, `gml`, `html`). Global defaults for these can be defined via #159.

Global settings will be moved to `cfg.yml` (`server`, `auth`).

### LINKED ISSUE #159: redesign store defaults + overrides
### default values

Currently default values for entity configurations (`services`, `providers`, `codelists`) are defined in the code via immutables annotations (`@Value.Default`). This approach has some drawbacks, mainly that a user cannot change the defaults, but also some issues with redundancy and merging.
The new approach will still define defaults in code (in another way), but it will also make the default values configurable by the same means as the entity configurations (`filesystem`, `rest`, `manager`). In the filesystem the default values will reside in one or multiple directory structures, the default parent directory will be `store/defaults`. There will be options to define all default values for an entity type in a single file or partition it into multiple files and subdirectories. For examples of such directory structures, see below.

### overrides

Currently overrides for an entity reside in `store/entities/entityType/#overrides#/entityId.yml`. The path is already configurable in the current implementation, but the default path will be changed to `store/overrides/entityType/entityId.yml`. The same options as for default values regarding multiple such directory structures and partitioning of files will be introduced. For examples, see below.

We might take this a step further and support overrides for different environments. So the default path might be `store/env/(dev|testing|production)/entityType/entityId.yml`. The current environment and the according subdirectory would then be chosen via parameter, environment variable or `cfg.yml`.

### polymorphism

Entities might have subtypes. For example providers may have subtype `feature/sql` or `feature/wfs` and we might introduce new providers like `styles`. Services also have subtypes, even though currently there is only `ogc_api`.
These subtypes are defined via fields in the entity configuration, e.g. `providerType` and `featureProviderType`, and are used during deserialization.

For overrides, we can ignore this fact, as these reference the entity id, so the subtype was already defined through the main entity configuration.

For default values, we have to take this into regard, because the implementation needs to know which defaults it should apply to a certain entity (and on a technical level how to deserialize the default values).
So if a user provides default values that are shared among all provider subtypes, these might be defined in `store/defaults/providers.yml`. But if the user provides default values for a specific subtype, this subtype has to be part of the path. We might allow both filenames or directory structures, e.g. `store/defaults/providers.feature.sql.yml` or `store/defaults/providers/feature/sql.yml`. For more examples, see below.

### examples

Paths for the store directories are defined in `cfg.yml`. To simulate a common setup, let's assume we have an inspire configuration with `kita` and `umweltzonen` that is used by multiple clients. The shared default values and entity configuration reside in the `store` directory, the client specific default values and overrides reside in `user_data`. So we have four directories that are processed in this order:

1. `store/defaults`
2. `user_data/defaults`
3. `store/entities`
4. `user_data/overrides`

#### store/defaults

This directory contains the default structure that is auto-generated by ldproxy, The values themselves might be customized, but let's assume ldproxy prefers a flat single file structure:

```fs
|-- store
|   |-- defaults
|   |   |-- providers.feature.sql.yml
|   |   |-- service.ogc_api.yml
```

All file contents are stripped down examples. Let's assume `service.ogc_api.yml` contains the default configuration for the `html` building block:

```yml
api:
- buildingBlock: HTML
  enabled: true
  noIndexEnabled: true
  schemaOrgEnabled: true
  layout: COMPLEX_OBJECTS
```

#### user_data/defaults

This directory contains client specific defaults that will be applied to both `kita` and `umweltzonen`. A common example for this would be `metadata`. Here we use the support for deep structures and file partitioning. The idea behind file partitioning is to not allow an arbitraty partitioning, but one following the object structure in the `yml`. So as `service` entities with subtype `ogc_api` have an object field `metadata`, we can use a nested file `metadata.yml`:

```fs
|-- user_data
|   |-- defaults
|   |   |-- services
|   |   |   |-- ogc_api
|   |   |   |   |-- metadata.yml
```

```yml
contactName: Geoinformationszentum, Information und Technik Nordrhein-Westfalen
contactEmail: geoinformationszentrum@it.nrw.de
licenseName: Datenlizenz Deutschland - Namensnennung - Version 2.0
licenseUrl: https://www.govdata.de/dl-de/by-2-0
```

#### store/entities

Nothing to see here, implementation will not change.

```fs
|-- store
|   |-- entities
|   |   |-- services
|   |   |   |-- kita.yml
|   |   |   |-- umweltzonen.yml
|   |   |-- providers
|   |   |   |-- kita.yml
|   |   |   |-- umweltzonen.yml
```

#### user_data/overrides

This directory contains client specific overrides. 
A common example for this would be `connectionInfo`. For `kita` we use the partitioning approach, in that case the entity id is a directory. For `umweltzonen` we use the flat approach, so the entity id is the filename.
We also disable `html` support for `umweltzonen`. Here we might implement an alias approach to allow referencing the api building block configuration by name:

```fs
|-- user_data
|   |-- overrides
|   |   |-- providers
|   |   |   |-- kita
|   |   |   |   |-- connectionInfo.yml
|   |   |   |-- umweltzonen.yml
|   |   |-- services
|   |   |   |-- umweltzonen
|   |   |   |   |-- html.yml
```

**connectionInfo.yml**
```yml
host: db
database: kita
user: postgres
password: cG9zdGdyZXM=
```

**umweltzonen.yml**
```yml
connectionInfo:
  host: db
  database: umweltzonen
  user: postgres
  password: cG9zdGdyZXM=
```

**html.yml**
```yml
enabled: false
```

### service generation + profiles

The current implementation has a second approach for default values that is used for service generation (`OgcApiCapabilityExtension`). This approach knows the concept of profiles, which was introduced to use different default values in xtraproxy than in ldproxy.
This will be replaced/merged with the approach described above. Profiles will be implemented using the same approach as for html templates. That means downstream products may define their own default values.

### environment variables

Recent discussions about setting passwords brought up the topic of setting configuration values via environment variables. There would be two options to implements this:

1. Explicit substitution
Dropwizard already has an inbuilt mechanism for this which could be easily activated for `cfg.yml`. This could also be expanded to the store. The (arbitrary) environment variable name would be defined in the `yml` file and replaced on load:
```
password: ${LDPROXY_KITA_DB_PASSWORD}
```

2. Generic
Another approach would be to generally allow overrides via environment variable. The variable name would then contain a configurable prefix and the path of the configuration key, for example:
```
LDPROXY_PROVIDERS_KITA_CONNECTIONINFO_PASSWORD
```

### schema migrations

Entity configurations are automatically migrated when a new software version introduces a new configuration schema. The schema version is saved in the configuration file using the field `entityStorageVersion`.
Default values and overrides might also need to be migrated. It has to be analyzed if every file needs to contain `entityStorageVersion` as well, or if there is a simpler approach.

azahnen:
closes #158, closes #159


azahnen:
There is still an open issue, therefore using draft.

cportele:
When I start an instance I get several exceptions when ldproxy tries to process the Mac `.DS_Store` files in the directory. Maybe only read files with an extension `.yml`?

cportele:
Another issue is that the defaults currently do not really work. If I set HTML defaults (e.g. for the base maps, the privacy notice links, etc.) and then in a service change some settings (bit leave the defaults untouched) it uses the system defaults, not the deployment defaults. I guess the `mergeInto()`-functions need an update, too.

cportele:
Thanks, that fixed the problems.

I have another issue now. During the startup phase (tile seeding, bbox computation) I get exceptions like "CRS EPSG:4326 is not supported" and "CRS EPSG:25832 is not supported" that I did not get before. However, I don't see how this could be related to the changes. Is Geotools just not yet ready? If yes, how can we make the startup more robust? Add loops to wait until Geotools is ready?

azahnen:
The original open issue is fixed as well.

azahnen:
I could not reproduce the CRS issue with various configurations. Can you send me the configuration that produces the issue?

PR #202, opened 2020-08-07
sort keys and constants

### LINKED ISSUE #192: Remove requirement for a column "id"
Currently the PGIS feature provider expects that each table has a column `id` with a unique integer.

One option would be to make this the default, but allow to specify in the configuration another column as the identifier column and also allow that the identifier may be a string, not an integer.

azahnen:
Fixes #192. 

### sort keys

Every table with mapped columns needs a sort key. Until now it was expected that every table has a numeric column with the same name that is used as sort key. The default sort key column name was already configurable:

```yml
connectionInfo:
  pathSyntax:
    defaultSortKey: id
```

This PR introduces support for string columns as sort key. Additionally it allows to set a different sort key per table:

```yml
managementrestrictionorregulationzone:
    sourcePath: /am_managementrestrictionorregulationzone{sortKey=inspireid}
    type: OBJECT
    properties:
      legislationCitation:
        sourcePath: '[id=id_managementrestrictionorregulationzone]am_manrestreg_legcit/[id_legislationcitation=id]am_legislationcitation{sortKey=identificationnumber}'
        type: OBJECT_ARRAY
```

Only the table at the end of the path should need a sort key, because that is the one with mapped columns.

### constants

This PR introduces support for constant values instead of column mappings.

```yml
name:
#  sourcePath: name
  constantValue: FOO
  type: STRING
```

When using `constantValue`, you don't need to define `sourcePath`.

### DIFF
--- a/xtraplatform.gradle
+++ b/xtraplatform.gradle
@@ -7,7 +7,7 @@ dependencies {
     }
     feature group: 'de.interactive_instruments', name: 'xtraplatform-store', version: '2.2.13'
     feature group: 'de.interactive_instruments', name: 'xtraplatform-web', version: '2.2.4'
-    feature group: 'de.interactive_instruments', name: 'xtraplatform-sdi-tools', version: '3.3.4'
+    feature group: 'de.interactive_instruments', name: 'xtraplatform-sdi-tools', version: '3.3.5'
 }
 
 allprojects {
--- a/xtraplatform.gradle
+++ b/xtraplatform.gradle
@@ -7,7 +7,7 @@ dependencies {
     }
     feature group: 'de.interactive_instruments', name: 'xtraplatform-store', version: '2.2.13'
     feature group: 'de.interactive_instruments', name: 'xtraplatform-web', version: '2.2.4'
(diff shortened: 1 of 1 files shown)


cportele:
The defaultSortKey change works in my tests, but I have a problem with constantValue. Here is the schema for an attribute:

```yaml
      serviceType:
        type: OBJECT
        objectType: Link
        label: Art des Dienstes
        description: Art des staatlichen Verwaltungsdienstes.
        properties:
          title:
            constantValue: Kinderbetreuung
            type: STRING
            label: Beschreibung
          href:
            constantValue: 'http://inspire.ec.europa.eu/codelist/ServiceTypeValue/childCareService'
            type: STRING
            label: Verweis
```

Only title is in the instance data:

```json
{
  "name": "Schwelm",
  "serviceType.title": "Brandschutzdienst",
  "pointOfContact.address.thoroughfare": "Hauptstr."
}
```

azahnen:
Please try again.

cportele:
That fixed the issue.

PR #216, opened 2020-09-07
More updates after testing

cportele:


azahnen (approved):
Just one comment/todo for after 2.0.0: 

The WFS provider detection via string comparison should be replaced with something more robust. For cases like this [FeatureProvider2](https://github.com/interactive-instruments/XtraPlatform-SDI-Tools/blob/master/xtraplatform-features/src/main/java/de/ii/xtraplatform/features/domain/FeatureProvider2.java) has `supports*()` methods. So we should introduce something like `supportsHighLoad()` that returns `false` for WFS providers.

cportele:
I will create an issue for this.

PR #257, opened 2020-09-30
add changelog

cportele:


.github_changelog_generator
```diff
@@ -0,0 +1,4 @@
+unreleased=false
+since-tag=v2.0.0
+pull-requests=false
+issues-wo-labels=false
```
azahnen, line 4: I would exclude any prereleases as these are mainly for internal usage and make the changelog much harder to read. According to the docs something like this should work:

```
exclude-tags-regex=([^v].*|v[^0-9].*|.*-.*)
```

This would also exclude any potential non-version tags.

I would also add two additional sections (not necessarily in this PR as there are currently no matching issues):

1. Dependency updates: self-explanatory 
2. Improvements: this would be for any non-functional changes to the software that are also no bugfixes. So for example the planned logging changes or performance optimizations. 

```
add-sections={"dependencies":{"prefix":"**Dependency updates:**","labels":["dependencies"]}, "improvement":{"prefix":"**Improvements:**","labels":["improvement"]}}
```

The categories would then make the semantic versioning straightforward. Any issue with `breaking` triggers a major release, any issue with `enhancement` triggers a minor release and anything else triggers a patch release.

azahnen (changes requested):

.github_changelog_generator
cportele: The behaviour of `exclude-tags-regex` is a bit surprising. Yes, the tags are excluded, but also all of the issues closed in the release, so the change log does not really reflect all the changes between the release versions. I had a look and did not find a way to get around this - or find another changelog generator that does it. So I would propose to accept that the pre-releases are still in the list.

`exclude-tags-regex` is still useful though to suppress version 1 tags.

cportele:
replaced by #297

PR #258, opened 2020-10-02
add query parameter skipGeometry

### LINKED ISSUE #223: add option to drop geometries in feature responses
via an additional query parameter, e.g. `geometry=false` or `returnGeometry=false`. Should be part of the building block "projections".

azahnen:
Closes #223


cportele (changes requested):
See the comment on the tile resource. 

(Also: In the v1 implementation the response does not contain a "geometry" property in the feature. This makes the response an invalid GeoJSON feature. If no geometry is returned, "geometry" must be `null`. In v2 the "geometry" property is there.)

azahnen:
v1 GeoJSON was fixed.

cportele (approved):

PR #277, opened 2020-10-13
Operational improvements

### LINKED ISSUE #265: deadlock in sql queries with small maxThreads
With `maxThreads=1`, a single request causes the deadlock. With higher values it seems to take the same number of parallel requests that are started nearly simultaneously  to trigger it.

If possible we should of course eliminate the deadlocks altogether. If that is somehow not possible or the solution takes longer to implement, we should at least forbid values for `maxThreads` that trigger the deadlock easily, maybe everything less than `4`.

### LINKED ISSUE #269: improve logging context
Use slf4js mdc feature to add additional columns with context information to the log:

1. **service id**
  Most actions are triggered by a service, be it through a request or as a background task. Therefore we should add a column with the service id.
2. **request/thread id**
  Thread ids could already be shown in the log, but they are of limited help because ldproxy uses a reactive architecture. Therefore most of the work does not happen on the request thread. 
  So we would need an additional column that shows either the thread id of the request or a generated request id even for messages from another thread.
  We should also make sure that messages are properly framed by start and end of request messages.

azahnen:
Fixes #265 
Closes #269


cportele (approved):
Tested maxConnections and reviewed log changes. Two comments:

* Can we also add the API context for background/startup tasks?
* CI fails, because the snapshot libraries are not found

azahnen:
I added the background task context.

I will merge this now so that we can proceed with testing, let's discuss the CI issue later.

PR #279, opened 2020-10-19
Operational improvements 2

### LINKED ISSUE #262: optimize service background task handling
Currently service background tasks like tile generation run in one thread per service. This can very quickly lead to a situation where the server load gets so high that no external requests can be handled.

The original design was to queue background tasks to the global scheduler, so that only one global thread is used to process the tasks one by one. This would implement the original design with an additional option to configure the number of parallel tasks that can be processed.

azahnen:
Closes #262 
Fixes issues with string sort keys in the provider and with request id logging.


cportele (approved):
I have pushed a few edits to the seeding class, please check.

In addition, I added changes to fix #280.

azahnen:
Looks good to me, I will merge this.

PR #287, opened 2020-10-26
fix api definition caching

azahnen:
The api definition caching did not take into account that configurations may change at runtime. I changed the cache key from the id to a hashsum to address this.


cportele (approved):
Looks good to me. I have tested it switching multiple building blocks on/off and off/on.

PR #290, opened 2020-10-27
Fix error 404 for features with a dot symbol in the ID

ulugbek:
Feature ID URI pattern now allows a dot symbol to avoid 404 errors for features with a dot symbol


cportele (approved):
I have simplified the regular expression and added an issue for the change (#291).

PR #297, opened 2020-10-27
generate changelog + create github release with gren

azahnen:
This is a replacement for #257.
I found another tool which seems a bit more robust and can also automatically create Github Releases: https://github.com/github-tools/github-release-notes

This PR contains the tool configuration, the current changelog file and the CI automation. In the future, when a tag is pushed, the CI will automatically create a Github Release with the corresponding changelog and push a new branch with the updated changelog file.
For the current release I executed these steps manually. I also updated Github Release 2.1.0 with the corresponding changelog.


cportele (commented):
That looks good - except for listing of v1.x.x up to v1.3.8, but not the later versions. I would drop all v1.x.x in the change log, since we do not really have proper release notes or the list of changes anyhow.

I will close the other PR.

PR #311, opened 2020-11-12
vector tile and style improvements/enhancements/bugfixes

### LINKED ISSUE #309: return CORS headers also for "Sec-Fetch-Mode" headers with value "cors"
It seems as if Chrome expects that behaviour, also without an explicit "Origin" header. [Sec-Fetch-Mode at MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-Fetch-Mode).

### LINKED ISSUE #310: various improvements to vector tiles and styles
The following improvements in the support for vector tile and style resources have been identified:
* support hashes in the URLs of the Mapbox GL web maps
* allow non-word characters in style identifiers
* extract style title from stylesheets
* make repairing polygons more robust, avoid errors if that fails and just ignore the geometry; if a merged geometry is invalid, use the un-merged features
* reduce number of vertices in vector tile geometries using JTS simplifier
* log errors when building the tile transformation context

(This issue originally covered also other topics, which now have been moved to separate issues.)

cportele:
Closes #310, closes #309


azahnen (approved):
I reviewed the code, I have only two suggestions:

- In `TilesConfiguration.mergeInto` you should add a block for `getRules()`. That makes the merging behaviour at least consistent and predictable, the most specific definition will overwrite all others. 
  In general this is a bit of an open issue. We need to decide where deep merging should be possible and where not and make the merging behaviour clear in the documentation.
- `FeatureTransformerTilesMVT` is getting huge, it would be good to factor out the separate concerns, e.g. a `TileGeometryProcessor` for the prepare and repair functionality and a `MvtFeatureMerger` for the new merge functionality.
   This is not critical for merging the pr.

I also tried out the instance and the web map, looks pretty good, no issues occurred.

PR #312, opened 2020-11-21
Manager

### LINKED ISSUE #264: advanced database connection configuration
Slick provides a lot of advanced options, e.g. for pool configuration. We should pass through all/most of these options from `connectionInfo` to slick.

### LINKED ISSUE #266: add admin task to reload entity configuration
A task under the admin endpoint at 7081 should be implemented that allows to reload all files from the store that are relevant for a certain entity and performs a rolling update. 
If an error occurs when reloading the files, it should be logged and/or returned from the task request, but the active entity stays as is.

### LINKED ISSUE #267: support auto-reload of entity configurations
Support a `watch` option for the store, that will trigger a rolling update of the affected entities when a file changes. See also #266.

### LINKED ISSUE #418: connectionInfo configuration is not merged

### LINKED ISSUE #419: reintroduce manager web app for easy configuration

azahnen:
Closes interactive-instruments/ldproxy-project-management#47.
Closes #264.
Closes #266.
Closes #267.
Closes #418.
Closes #419.
 

To test this PR you can run:
```
docker run --name ldproxy -d -p 7080:7080 -v ~/your_data_dir:/ldproxy/data docker.ci.interactive-instruments.de/iide/ldproxy:manager2
```
The manager will start at `http://localhost:7080/manager/`.

If you want to use a database on localhost, use the special hostname `host.docker.internal` to make it work inside the docker container (see [here](https://docs.docker.com/docker-for-mac/networking/#use-cases-and-workarounds) or [here](https://docs.docker.com/docker-for-windows/networking/#use-cases-and-workarounds)). 

Another option is to use the instance at: http://ldproxy01/manager2/manager/ (login with admin/ldproxy)

#### Work in Progress

- [x] backend sync
- [x] security
- [x] view codelists
- [x] view connection info
- [x] i18n + help

#### Not planned for this PR

- add|edit|delete codelists
- view|edit|delete providers
- overrides


cportele:
Problems:

- [ ] Text-valued properties that are cleared are not removed. Example Services > Metadata > Contact URL first set to http://example.com/ and then removed again. The entry was correctly set in the first edit, but not removed in the configuration file in the second edit.
- [ ] Services > Collections > Data. I have disabled the "queryable" aspect of a queryable property - no change in the configuration file. The reenabled the "queryable" aspect again - the property is again a queryable, but all other queryables are no longer queryable.

I will add more issues here that I notice during tests.

cportele:
New problems:

* [ ] When creating a new service with the umweltzonen DB, the native CRS is reported as 4326 LON_LAT, but the data is in 25832. It is also not possible to specify the native CRS during the creation of a new API. The CRS detection works with the vineyards DB.
* [ ] Tab "Api". With Features HTML, Features GeoJSON and Coordinate Reference Systems all disabled, Features Core is still greyed out, because it is "required by Coordinate Reference Systems".
* [ ] If OpenAPI 3.0 is disabled, the API becomes unusable. (I will have a look at this problem.)
* [ ] If I select a property and switch the queryable status, the change only takes effect, if I wait long enough for the change to be written before doing anything else in the UI. If it has to be that way, then the UI should not accept and clicks until the change has been written.
* [ ] I disabled Features HTML, Features GeoJSON, HTML and JSON and later enabled them again. The service config looks correct after these steps, but the API still did not accept HTML for features, only JSON. I have not been able yet to understand why and how that happens.
* [ ] More of the question: It is not possible to change to provider to use another property as the ID and one has to do it in YAML file?

azahnen:
> * [ ]  When creating a new service with the umweltzonen DB, the native CRS is reported as 4326 LON_LAT, but the data is in 25832. It is also not possible to specify the native CRS during the creation of a new API. The CRS detection works with the vineyards DB.

That is actually the intended behaviour as in this case the srid for column `geometry` in table `am_managementrestrictionorregulationzone` is set to `0` in the db. In that case we fall back to CRS84.

> * [ ]  Tab "Api". With Features HTML, Features GeoJSON and Coordinate Reference Systems all disabled, Features Core is still greyed out, because it is "required by Coordinate Reference Systems".

It looks like CRS is not really disabled, I will look into it.

> * [ ]  If I select a property and switch the queryable status, the change only takes effect, if I wait long enough for the change to be written before doing anything else in the UI. If it has to be that way, then the UI should not accept and clicks until the change has been written.

That is a bug, I will look into it.

> * [ ]  I disabled Features HTML, Features GeoJSON, HTML and JSON and later enabled them again. The service config looks correct after these steps, but the API still did not accept HTML for features, only JSON. I have not been able yet to understand why and how that happens.

I am getting a NullPointerException when reproducing your steps, it even persists after a restart. It seems that the HTML configuration is not correctly merged. I will look into it.

> * [ ]  More of the question: It is not possible to change to provider to use another property as the ID and one has to do it in YAML file?

No, that is not possible as mentioned at the top:

> #### Not planned for this PR
> - add|edit|delete codelists
>- view|edit|delete providers

azahnen:
@cportele All known issues should be resolved. I also pushed the tests and fixes for all `mergeInto` methods.

cportele:
Thanks @azahnen - will test again over the next days

cportele:
@azahnen - Another issue is that I cannot log in to the Manager. When the login page is loaded, I can immediately see `POST http://localhost:7080/rest/auth/token 400 (Bad Request)` in the console. Pressing "Login" has no effect (no new connection is recorded in the Network tab). I tried this with the new manager2 image and a local build in the IDE, same result.

azahnen:
@cportele The tiles issue should be fixed. Regarding the login issue, the initial 400 is expected. If you can't login, try to delete all cookies for localhost:7080. Also see https://github.com/interactive-instruments/ldproxy/blob/manager2/docs/de/configuration/global-configuration.md#autorisierung.

cportele:
@azahnen - Yes, the tiles issue is gone. The login issue however is still there. I have removed all cookies and also tried with an existing and a new data directory. Tried Chrome and Firefox. No change in the behaviour and also no warning is written to the log. Any other idea?

azahnen:
@cportele Please try again, the login should work now. There is still another issue where error messages are shown in the password change form before entering anything. Both bugs were introduced along with the changes to auto-submit forms on context changes but somehow did not occur in my local environment without a clean.

azahnen:
@cportele I fixed the remaining issue with form validation. I could not reproduce the login issue.

I updated the description at the top to use the docker image from the internal registry. The docker hub image will not be updated any more.

azahnen:
@cportele The docker image was broken due to a build issue. The manager jar did not contain the web app, therefore the 404.

I updated the description at the top with a link to the deployed and working instance.

I also found and fixed an issue with WFS based services that was introduced by the sort key implementation.

PR #329, opened 2020-11-25
Add temporal extent computations

### LINKED ISSUE #174: support temporal extent computation
Add support for computing the temporal extent of a collection from the data, like for spatial extents.

ulugbek:
Closes #174.


ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ldproxy/ogcapi/features/core/app/FeaturesCoreDataHydrator.java
```diff
@@ -241,4 +254,84 @@ private boolean hasMissingBboxes(Map<String, FeatureTypeConfigurationOgcApi> fea
...
+
+                        List<String> temporalQueryables = entry.getValue()
+                                .getExtension(FeaturesCoreConfiguration.class)
+                                .flatMap(FeaturesCoreConfiguration::getQueryables)
+                                .map(FeaturesCollectionQueryables::getTemporal)
+                                .orElse(ImmutableList.of());
+
+                        if (!temporalQueryables.isEmpty()) {
+                            Optional<Interval> interval;
+                            if (temporalQueryables.size() == 2) {
```
azahnen, line 295:
```suggestion
                            if (temporalQueryables.size() >= 2) {
```

azahnen (changes requested):

ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ldproxy/ogcapi/features/core/app/FeaturesCoreDataHydrator.java
```diff
@@ -241,4 +254,84 @@ private boolean hasMissingBboxes(Map<String, FeatureTypeConfigurationOgcApi> fea
...
+                                        .getTemporalExtent(entry.getValue().getId(), temporalQueryables.get(0));
+                            }
+
+                            if (!interval.isPresent()) {
+                                return entry;
+                            }
+
+                            TemporalExtent temporalExtent = new ImmutableTemporalExtent.Builder()
+                                    .start(interval.get().getStart().toEpochMilli())
+                                    .end(interval.get().getEnd().toEpochMilli())
```
azahnen, line 309: This line throws an exception when the end of the interval was set to `Instant.MAX` because the Sql-Query returned `null`. To fix it, only set the end of `TemporalExtent` if not `interval.isUnboundedEnd()`.

azahnen (changes requested):

azahnen (approved):

cportele (approved):
I made two changes:

* The configuration parameter names in the doc did not reflect the names in the code. While the ones in the doc were clearer, I kept the ones in the code because `spatialCompute` is in use.
* The documentation was change to say that the compute flags could also be set in `defaultExtent`, however, this was not implemented. I made the necessary changes.

@azahnen - please have a look at the changes, otherwise this is ready to be merged

azahnen:
@cportele I reverted your changes to `FeaturesCoreDataHydrator` and instead apply the `defaultExtent` in `OgcApiDataV2.mergeCollectionDefaults`.

I am wondering if we should enable computations by default in `defaultExtent`, what do you think?

cportele:
> I am wondering if we should enable computations by default in defaultExtent, what do you think?

@azahnen - I thought about that, too, but decided against it because it changes the default behaviour. So far, computation has always been activated with a `true` and `false` is never used AFAIK. However, computation by default should be ok for most APIs and is a good practice in general, but there are a few that would have to be updated to explicitly add a `false`. @rkoblenzer - do you have any concerns about such a change or would it be ok for you?

azahnen:
I think for spatial extents it would only change the behaviour if neither a default extent nor a collection extent is set.

azahnen:
At least that would be the intended behaviour, I think the code currently does not completely reflect that. An explicit default spatial extent should have preference over the default spatial compute flag.
A collection spatial extent would always completely override the default spatial exent and its compute flags are false be default.

azahnen:
To make it a bit clearer, in both of these cases the explicit extent would win, nothing is computed.

```yml
id: lika
defaultExtent:
  spatial:
    xmin: 5.61272621360749
    ymin: 50.2373512077239
    xmax: 9.58963433710139
    ymax: 52.5286304537795
collections:
  verwaltungseinheit:
    extent: null
```

```yml
id: lika
defaultExtent: null
collections:
  verwaltungseinheit:
    extent:      
      spatial:
        xmin: 5.61272621360749
        ymin: 50.2373512077239
        xmax: 9.58963433710139
        ymax: 52.5286304537795
```

cportele:
I have a different view:

* The default extent is just a default and should be overridden by any extent declaration on the collection level. So, with a default extent, if compute is active for a collection, then the extent should be computed.
* There is use in specifying an explicit extent and computing it, too. The explicit extent (on the API or collection level) is a start value that is immediately present when the API is started, but it will be replaced by the computed extent, once it has been computed (which may take a while).

azahnen:
> The default extent is just a default and should be overridden by any extent declaration on the collection level. So, with a default extent, if compute is active for a collection, then the extent should be computed.

No disagreement here, that's what I tried to say.

> There is use in specifying an explicit extent and computing it, too. The explicit extent (on the API or collection level) is a start value that is immediately present when the API is started, but it will be replaced by the computed extent, once it has been computed (which may take a while).

That is not how it currently works, the API only gets available when computations are finished.

cportele:
> That is not how it currently works, the API only gets available when computations are finished.

Right, and I do not see that we are going to change that now. 

So the rule would be that the (spatial/temporal) extent of a collection is computed, if and only if the following conditions are matched:

* the feature provider supports the computation
* the collection has a queryable
* there is no explicit extent defined for the collection
* the compute flag is 
  * not set to `false` in the collection or 
  * the flag is not set in the collection and set to `false` in the default extent

If the extent is not computed, it will use the extent in the collection, if one is set, or from the default extent, if that has a value, or no extent otherwise.

Correct?

azahnen:
> the collection has a queryable

I don't think this is currently taken into account for spatial extents, but otherwise that should describe the current behaviour.

The use case I had in mind with the changed default value was mainly the auto mode, so that a newly created service has enabled computations. But we actually don't have to change the defaults for that, enabling spatial extent computation is already part of the auto mode.

So to sum this up, we will not change the default values now, but I would enable temporal extent computation in auto mode, as we already do for spatial extents.

azahnen:
@cportele Any other thoughts, or can I merge this?

cportele:
@azahnen - I have updated the documentation to bring it inline with what we have now. Have look and merge, if everything is fine.

cportele:
> I reverted your changes to FeaturesCoreDataHydrator and instead apply the defaultExtent in OgcApiDataV2.mergeCollectionDefaults.

@azahnen - I just noticed that the logic in `mergeCollectionDefaults` is not what I expected. It is either use the whole extent or use the whole defaultExtent, but this does not support that I can handle spatial and temporal aspects differently. The merge logic should merge spatial and temporal parameters separately. I will create another PR to correct that.

Issue #263, opened 2020-10-06
support ogcapi extension startup/health checks


azahnen:
The ogcapi service should ask every extension on startup if it has all the resources it needs, for example the feature core extension needs a feature provider. If the feature provider is not there, startup should fail. 

A related topic that should be considered is service hydration. Service hydration takes place before startup tasks and also delegates to extensions. Currently, if no feature provider is found and for example spatial extents should be computed, hydration fails but service seems to start normally.

So at least error handling during hydration has to be improved, but most likely startup checks should happen before or as part of the hydration.

We should also verify if it is easy to implement the same checks as dropwizard health checks. That would allow continuous monitoring of the service state at the admin endpoint at 7081.

cportele:
closed by #362, open issues covered in #376, #377, #378, #379

PR #330, opened 2020-11-27
mapping validation

### LINKED ISSUE #260: add optional mapping validation on startup
The provider should check on startup if the mapping and the database fit together. 

This check should be optional, meaning it can be enabled/disabled in the provider configuration. Additionally I would provide a configuration option to decide if a failed validation throws an error or just logs a warning.

azahnen:
Closes #260. 
Closes interactive-instruments/ldproxy-project-management#55.


cportele:
I tried this by adding `ignoreTypes: WARN` to the lika provider configuration. Without this, ldproxy starts as usual. With this, I get the following:

```log
DEBUG [2020-11-27 18:30:04,319]                          - Listening for events for [defaults]  [de.ii.xtraplatform.store.app.EventSourcing] [stream.events-5]  
DEBUG [2020-11-27 18:30:04,319]                          - Replaying events for entities  [de.ii.xtraplatform.store.app.EventSourcing] [stream.events-6]  
DEBUG [2020-11-27 18:30:04,434]                          - Replaying events for overrides  [de.ii.xtraplatform.store.app.EventSourcing] [stream.events-5]  
DEBUG [2020-11-27 18:30:05,330]                          - Registered entity: class de.ii.xtraplatform.feature.provider.sql.app.FeatureProviderSql lika  [d.i.x.store.app.entities.EntityRegistryImpl] [startup]  
DEBUG [2020-11-27 18:30:05,330]                          - Entity created: Identifier{id=lika, path=[providers]}  [d.i.x.store.app.entities.EntityDataStoreImpl] [startup]  
DEBUG [2020-11-27 18:30:05,587]                          - Deregistered entity: class de.ii.xtraplatform.feature.provider.sql.app.FeatureProviderSql lika  [d.i.x.store.app.entities.EntityRegistryImpl] [startup]  
DEBUG [2020-11-27 18:30:05,587]                          - ENTITY REMOVED lika  [d.i.x.store.app.entities.EntityRegistryImpl] [startup]  
DEBUG [2020-11-27 18:30:05,606]                          - Registered entity: class de.ii.ldproxy.ogcapi.app.OgcApiEntity lika  [d.i.x.store.app.entities.EntityRegistryImpl] [startup]  
ERROR [2020-11-27 18:30:05,607]                          - E  [d.i.x.store.app.entities.EntityRegistryImpl] [startup]  
! java.lang.IllegalStateException: No feature provider found.
! at de.ii.ldproxy.ogcapi.features.core.app.FeaturesCoreProvidersImpl.lambda$getFeatureProvider$0(FeaturesCoreProvidersImpl.java:43)
! at java.base/java.util.Optional.orElseThrow(Optional.java:408)
! at de.ii.ldproxy.ogcapi.features.core.app.FeaturesCoreProvidersImpl.__M_getFeatureProvider(FeaturesCoreProvidersImpl.java:43)
! at de.ii.ldproxy.ogcapi.features.core.app.FeaturesCoreProvidersImpl.getFeatureProvider(FeaturesCoreProvidersImpl.java)
! at de.ii.ldproxy.ogcapi.features.core.domain.FeaturesCoreProviders$$Proxy.getFeatureProvider(Unknown Source)
! at de.ii.ldproxy.ogcapi.tiles.VectorTileSeeding.__M_isEnabledForApi(VectorTileSeeding.java:100)
! at de.ii.ldproxy.ogcapi.tiles.VectorTileSeeding.isEnabledForApi(VectorTileSeeding.java)
...
! at java.base/java.lang.Thread.run(Thread.java:834)
INFO  [2020-11-27 18:30:05,612]                     lika - Service with id 'lika' started successfully.  [d.i.x.services.domain.AbstractService] [startup]  
DEBUG [2020-11-27 18:30:10,336]                     lika - Entity created: Identifier{id=lika, path=[services]}  [d.i.x.store.app.entities.EntityDataStoreImpl] [pool-3-thread-1]  
DEBUG [2020-11-27 18:30:10,337]                     lika - Listening for events for [entities, overrides]  [de.ii.xtraplatform.store.app.EventSourcing] [pool-3-thread-1]  
```

The API is there, but any request results in a 500 response (which is just logged as DEBUG and with no details about the error):

```
DEBUG [2020-11-27 18:34:20,017]                     lika - Processing request: GET / [1484d666-3ae0-4007-a128-4601b84ad39f] [d.i.x.services.app.ServicesEndpoint] [dw-114]  
DEBUG [2020-11-27 18:34:20,139]                     lika - Sending response: 500 Internal Server Error [1484d666-3ae0-4007-a128-4601b84ad39f] [d.i.x.services.app.LoggingContextCloser] [dw-114]  
```

cportele:
I added `validateTypes: ERROR` to the vineyards configuration and I get

```
ERROR [2020-11-28 09:05:19,990]                vineyards - Invalid role ID for property 'registerId' in type 'vineyards': column 'wlg_nr' in table 'vineyards' should not be used as feature id, neither a primary key nor a unique constraint can be found  [d.i.x.f.domain.AbstractFeatureProvider] [startup]  
```

However, the provider and the service starts and the API can be used as usual. This is not consistent with the documentation: 

> bei `ERROR` werden Fehler geloggt und der Provider wird nicht gestartet

cportele:
I changed a DATETIME property to INTEGER in the daraa configuration for one feature type, but this type mismatch was not reported. Shouldn't such cases be checked, too?

(Also, no error was reported that an "integer" is used as a temporal queryable.)

cportele (changes requested):
See the comments to this PR.

azahnen:
> I tried this by adding ignoreTypes: WARN to the lika provider configuration. Without this, ldproxy starts as usual. With this, I get the following:



> I added validateTypes: ERROR to the vineyards configuration and I get
However, the provider and the service starts and the API can be used as usual.



> I changed a DATETIME property to INTEGER in the daraa configuration for one feature type, but this type mismatch was not reported. Shouldn't such cases be checked, too?

azahnen:
Sorry, I clicked on `Close with comment` by accident. 
As for your issues:

> I tried this by adding ignoreTypes: WARN to the lika provider configuration. Without this, ldproxy starts as usual. With this, I get the following:

The schema crawler did not correctly decode the passwords. Not sure why it worked in other cases but I fixed it.
The lika configuration was also a good test because validation took 5 minutes. I reduced that to 30 seconds by only crawling the schema for tables that are actually used in the mapping. (The validation reused the code from auto mode which has to crawl all tables.)
I also made the progress more visible by adding an `INFO` message for each type and also by splitting up the validation by type. So instead of seeing the warnings/errors only when everything is done, they are printed after each type.

> I added validateTypes: ERROR to the vineyards configuration and I get
However, the provider and the service starts and the API can be used as usual.

This was a general bug regarding entity lifecycles, so it might have affected other startup errors in providers or services as well. I fixed it.
Note that only the provider will not start, the service starts with the usual no feature provider found message. Fixing that would be part of #263.

> I changed a DATETIME property to INTEGER in the daraa configuration for one feature type, but this type mismatch was not reported. Shouldn't such cases be checked, too?

I did not test this case yet, but the check is implemented and should fail in this case, yes. I will test it later today.

> (Also, no error was reported that an "integer" is used as a temporal queryable.)

This is a provider startup check, it doesn't know anything about queryables. Validating queryables should also be part of #263.

azahnen:
> I changed a DATETIME property to INTEGER in the daraa configuration for one feature type, but this type mismatch was not reported. Shouldn't such cases be checked, too?

I misinterpreted this at first, I thought you mapped an integer column to a DATETIME property, but you switched the property type. There is currently no test for INTEGER properties, the implemented tests are here: https://github.com/interactive-instruments/ldproxy/issues/260#issuecomment-717951853.

I agree that in general we should have tests for all types, but that means we first have to define the possible mappings. 
So for each property type, we have to define the column types that

1. can always be mapped, regardless of the data (`timestamp -> DATETIME`)
2. can be mapped if the data fulfills certain requirements (`string -> DATETIME`)
3. cannot be mapped (`boolean -> DATETIME`)

Then we can implement the tests accordingly, but we might also implement some additional mappings. For example if we allow `string -> DATETIME` there is no reason to forbid `integer -> DATETIME` for unix timestamps.

I would then suggest to change the configuration parameter to `typeValidation: NONE|LAX|STRICT`. 
`LAX` would only fail for non-existing tables or columns and for type 3 mappings. It would warn about cases that might work but cannot be guaranteed to work, like type 2 mappings and sort keys or feature ids without uniqueness constraint.
`STRICT` would also fail for the cases that are warnings in `LAX`. So in `STRICT` mode the provider only starts when flawless operation does not depend on the data.

azahnen:
I did not implement additional value type checks yet, I will create a follow-up issue for that.
I changed the configuration parameter to `typeValidation: NONE|LAX|STRICT` and adjusted the docs accordingly.

cportele (approved):

PR #362, opened 2021-02-07
Improvements

### LINKED ISSUE #353: inconsistencies in getPathPattern()
A handful of concrete format extension classes override the getPathPattern() method, which is in most cases unnecessary und leads to inconsistencies, if the path pattern evolves, but not all subtypes are updated.

Fix: Remove all unnecessary overrides.

### LINKED ISSUE #356: Make property references more tolerant/robust
Property references are used in the service configuration in several places, e.g., in the list of queryables or as keys in a transformations map. Sometimes square brackets are used to identify arrays and sometimes not. The idea to use property references always without square brackets. But it should still work, if square brackets are present for compatibility with existing configurations.

Also, property references should be validated and those that do not match the provider schema should be ignored and reported as warnings in the log. This extends #350.

cportele:
* validate all property references (queryables and transformation keys) and log/remove invalid ones during hydration
* validate the types of queryables and remove queryables that have been classified incorrectly
* normalise all property references in the service configuration to the variant without square brackets during hydration
* Detect missing feature providers and feature types in the provider during hydration and remove collections without a provider
* fix inconsistent path patterns for `collectionId` / `featureId`
* update dependency to the vector tiles library

This PR closes #353, closes #356. It also is a start to address #263.

Tested with all current ldproxy.net deployments.


azahnen:
After reviewing the changes I think this needs a more detailed discussion. I am not sure if the approach of removing elements with invalid references from the configuration and starting the service anyway is very helpful.

azahnen:
To implement the strict validation and prevent the service startup, you would have to override `onStartup` in `OgcApiEntity`. You can see how it is implemented in the provider here: https://github.com/interactive-instruments/xtraplatform-spatial/blob/250743e4a4f114019b08a919d29f8538910e953e/xtraplatform-features/src/main/java/de/ii/xtraplatform/features/domain/AbstractFeatureProvider.java#L80

cportele:
* `apiValidation` with values `NONE`, `LAX` and `STRICT` has been implemented.
* Checks have been added for all stable/draft modules.
* Invalid transformations and queryables are no longer removed during hydration.

This also closes #263. However, there are a few open questions:

* Currently StartupResult has been created based on ValidationResult. ValidationResult and the VALIDATION enum should be decoupled from the type validation and so that they can be used directly in the API validation.
* In which extension should we execute the startup checks for the configuration of a module? For most modules I have added the checks in the key components of the module (endpoint, format, parameter), but in Tiles this is not straightforward and I have added the configuration check in the Capability component.

azahnen:
Regarding the open questions: 

> Currently StartupResult has been created based on ValidationResult. ValidationResult and the VALIDATION enum should be decoupled from the type validation and so that they can be used directly in the API validation.

I moved ValidationResult and VALIDATION to xtraplatform-core and replaced usages of StartupResult with ValidationResult.

> In which extension should we execute the startup checks for the configuration of a module? For most modules I have added the checks in the key components of the module (endpoint, format, parameter), but in Tiles this is not straightforward and I have added the configuration check in the Capability component.

I think the Capability component (which should be renamed to BuildingBlock following the interface) is the right place. It was intended to be kind of the main extension for a module and also handles other configuration related tasks like providing the default configuration. For further discussion, see TBD.

> Should we also add a check in each component that all imported components (via `@Requires`) are available?

As already discussed this is more of a technical issue with ipojo and should not be part of our startup checks (see #376).

azahnen (approved):
In general the code looks good and works as expected.

As discussed I am still not totally satisfied with the handling of square brackets in property references, but I think it is fine for now. 

I created two follow-up issues to define a process for configuration schema upgrades (#378) and to clarify the boundaries between migration, hydration and validation (#379).

PR #367, opened 2021-02-12
support disabling OAS30

### LINKED ISSUE #366: exception when OAS30 is disabled
If OpenAPI 3.0 is disabled, an exception is thrown. There are two bugs:
* The path parameter `resource` is part of the OAS30 module, not the Common module, but the resource is defined in the Endpoint in Common. The class needs to be moved to common.
* `isApplicable()` in the class had an incorrect check on the `definitionPath`.

Disabling OAS30 will always result in an invalid API since at the moment and should not be considered, because OpenAPI 3.0 is the only type of API Definition supported by ldproxy. Some parts of the code assume that OAS30 is always enabled and this should be changed in this update, too, because in the future other API Definition options will be added; for example, OpenAPI 3.1.

cportele:
closes #366


azahnen (approved):
Works as expected.

PR #370, opened 2021-02-22
add support for OGC API Records

### LINKED ISSUE #369: Add support for OGC API Records
Current draft: http://docs.opengeospatial.org/DRAFTS/20-004.html

Changes:
- support parameter `q`
- use itemType `record` to identify record collections (catalogues)
- add startup checks to verify that a record collection supports the requirements (queryables to search with parameter `q`, queryables `type`, `externalid`, etc.)
- update conformance declaration (add Common, Records)
- support collectionId path parameter per API (`collectionIdAsParameter`)
- if all collection definitions are identical, support the use of one collection to derive schema and query parameters for all Items/Item resources (`collectionDefinitionsAreIdentical`)
- alphabetically sort enum values in OpenAPI definitions

Support for sorting may be added later (but it is a separate conformance class).

cportele:
closes #369 

Sample instance: https://github.com/opengeospatial/ogcapi-records/blob/master/implementations.md#ldproxy


azahnen (approved):
I played around with the deployment, looks good too me. 

The q implementation looks like I would have expected.

One thing I noticed are the crs and extent related properties which don't seem to make sense in this context:

- https://geoinfodok.ldproxy.net/geoinfodok/collections?f=json
- https://geoinfodok.ldproxy.net/geoinfodok/collections/aa_advstandardmodell?f=json

cportele:
Thanks @azahnen - Yes, we could drop `extent` and `crs`, if there are no spatial or temporal properties. So far, this is not done, also not for features, but I agree it would make sense. I will have a look, but I will create a new issue and PR for this.

In this deployment, the schema for the records is actually a more complete record schema (geometry is a core queryable in Records), which can also be seen in the [response schema](https://geoinfodok.ldproxy.net/geoinfodok/collections/aa_advstandardmodell/schema), but all the geometry properties for the records is set to null. In this case, we could drop the geometry from the schema and remove the geometry queryable, since no enum will ever have a geometry.

PR #385, opened 2021-03-02
Fixes and sort keys

### LINKED ISSUE #217: Improve provider analyses in the Vector Tiles module
The Vector Tiles module analyses the feature provider type and disables tiles support for those that are "WFS". This should replaced with something more robust that continues to work, when we add more providers. For cases like this FeatureProvider2 has `supportsXyz()` methods. So we should introduce something like `supportsHighLoad()` that returns false for WFS providers.

### LINKED ISSUE #371: default CRS in spatial extents is EPSG 4326, not CRS84
If a spatial extent is created in a service configuration (e.g., as a default extent) with just the coordinates, the CRS is set to EPSG 4326 without the lon/lat force flag. This creates errors in the new draft startup checks which expects CRS84 or CRS84h as the CRS of a collection extent.

### LINKED ISSUE #381: limit/offset ignored in some queries
In some cases the `limit`/`offset` values are ignored and all matching features are returned by FeatureSourceStream of the feature provider. This is at least the case with OR predicates.

Example: https://demo.ldproxy.net/vineyards/collections/vineyards/items?filter=subregion%20LIKE%20%27%25Bernkastel%25%27%20OR%20village%20LIKE%20%27%25Bernkastel%25%27%20OR%20searchfield2%20LIKE%20%27%25Bernkastel%25%27&limit=2

azahnen:
Fixes #381.
Fixes #371.
Closes #217.
Prepares for interactive-instruments/ldproxy-project-management#41.

### DIFF
--- a/docs/de/configuration/services/building-blocks/README.md
+++ b/docs/de/configuration/services/building-blocks/README.md
@@ -34,6 +34,7 @@ Grundsätzliche Regeln, die für alle API-Module gelten, finden Sie [hier](gener
 |[Filter / CQL](filter.md) |FILTER |draft |Nein |Aktiviert die Angabe von CQL-Filtern für die Ressourcen "Features" und "Vector Tiles"
 |[Geometry Simplification](geometry-simplification.md) |GEOMETRY_SIMPLIFICATION |draft |Nein |Aktiviert die Option zur Vereinfachung von Geometrien nach Douglas-Peucker bei den Ressourcen "Features" und "Feature"
 |[Projections](projections.md) |PROJECTIONS |draft |Nein |Aktiviert die Option zur Begrenzung der zurückgelieferten Feature-Eigenschaften bei den Ressourcen "Features", "Feature" und "Vector Tiles"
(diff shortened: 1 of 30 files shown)


cportele (changes requested):
This looks good, everything worked as expected in tests, except one problem. If some values of the sort field contain a single quote character, I get errors like these, because the quote is not escaped:

```
! org.postgresql.util.PSQLException: Unterminated string literal started at position 237 in SQL
SELECT A.name AS CSKEY_0, A.ogc_fid AS SKEY, A.ogc_fid, A.name, ST_AsText(ST_ForcePolygonCCW(A.geom)) FROM airports A WHERE (A.ogc_fid >= 1 AND A.ogc_fid <= 40 AND A.name >= 'Papa Stour Airstrip' AND A.name <= 'Wick John O'Groats Airport') ORDER BY 1 DESC,2.
```

https://demo.ldproxy.net/zoomstack/collections/airports/items?q=%27

azahnen:
@cportele The latest xtraplatform-spatial snapshot should fix the missing escaping.

cportele (approved):
Yes, that fixed the issue.

Issue #399, opened 2021-03-12
Support option to reduce the number of decimal places


cportele:
Too many decimal places reduce the readability and increase the data volume without any benefit to users. It should be possible to limit the number of digits for each axis depending on the CRS.

This affects mainly feature geometries, but also the generation of `bbox` values in the HTML filter editor.

PR #400, opened 2021-03-12
add coordinatePrecision option

cportele:
This PR starts to address #399. It adds the capability to specify the number of maximum number decimal digits in coordinates depending on the unit of the number. The default is still that no digits are cut. 

Example: 

```yaml
- buildingBlock: FEATURES_CORE
  coordinatePrecision:
    metre: 2
    degree: 7
``` 

Open issues:

- There can be different units in a CRS (in particular - horizontal is degrees, vertical is meters). Currently there is only one precision value for all axes.
- The HTML Filter editor also creates coordinates for the `bbox` parameter that have some 15 decimal places or so. Since this one is always in degrees and just a rough filter, maybe always restrict this to 4 decimal places (approx 11 meters) or so?
- Currently only meter and degree are supported as units. For the typical cases, this should be sufficient, at least for now. So maybe that is not something we need to address now.


azahnen (approved):
I moved the unit extraction to xtraplatform-spatial, geotools shouldn't be used directly.

Issue #376, opened 2021-02-24
replace all field injections with constructor injections


azahnen:
Field injections are generally considered an anti-pattern, as they make instantiating classes in a context without a dependency injection container much harder (e.g. in unit tests). With iPOJO there might also be issues with null injections when using field injections.

Therefore we should replace all field injections with constructor injections. A technical issue in this context (which is the reason why we used field injection in the first place in some cases) is that constructor injections are only allowed for interfaces. So if a class without interface is injected into a field, an interface has to be extracted before refactoring to constructor injection.

Tasks:

- [x] search for every occurrence of `@Requires` in ldproxy, xtraplatform-spatial and xtraplatform
- [x] if `@Requires` annotates a field and the field type is a class, extract an interface and use that instead (right click -> Refactor -> Extract Interface in Intellij)
- [x] replace field injection ([example](https://github.com/interactive-instruments/ldproxy/blob/ec68c9f8997da462cfa8b60f49207ba36bf7a7a6/ogcapi-draft/ogcapi-observation-processing/src/main/java/de/ii/ldproxy/ogcapi/observation_processing/application/ObservationProcessingOnLandingPage.java#L33)) with constructor injection ([example](https://github.com/interactive-instruments/ldproxy/blob/ec68c9f8997da462cfa8b60f49207ba36bf7a7a6/ogcapi-draft/ogcapi-observation-processing/src/main/java/de/ii/ldproxy/ogcapi/observation_processing/application/ObservationProcessingOnLandingPage.java#L39))

bekakh:
@azahnen - Should `@Context` field injections also be replaced by constructor injections?

cportele:
addressed by #401

PR #401, opened 2021-03-15
Replace field injection with constructor injection

bekakh:
Addresses https://github.com/interactive-instruments/ldproxy/issues/376, all field dependency injections were replaced by constructor injections.


ogcapi-stable/ogcapi-common/src/main/java/de/ii/ldproxy/ogcapi/common/app/QueriesHandlerCommon.java
```diff
@@ -28,14 +28,11 @@
 
 @Component
 @Instantiate
-@Provides(specifications = {QueriesHandlerCommon.class})
+@Provides
 public class QueriesHandlerCommon implements QueriesHandler<QueriesHandlerCommon.Query> {
```
azahnen, line 32:
```suggestion
public class QueriesHandlerCommonImpl implements QueriesHandlerCommon {

public interface QueriesHandlerCommon extends QueriesHandler<QueriesHandlerCommon.Query> {}
```

ogcapi-stable/ogcapi-common/src/main/java/de/ii/ldproxy/ogcapi/common/infra/EndpointConformance.java
```diff
@@ -39,11 +39,12 @@
 
     private static final List<String> TAGS = ImmutableList.of("Capabilities");
 
-    @Requires
-    private QueriesHandlerCommon queryHandler;
+    private final QueriesHandler<QueriesHandlerCommon.Query> queryHandler;
 
-    public EndpointConformance(@Requires ExtensionRegistry extensionRegistry) {
+    public EndpointConformance(@Requires ExtensionRegistry extensionRegistry,
+                               @Requires QueriesHandler<QueriesHandlerCommon.Query> queryHandler) {
```
azahnen, line 45:
```suggestion
                               @Requires QueriesHandlerCommon queryHandler) {
```

azahnen (changes requested):
I fixed most of the issues.

There is one common pattern left that does not work and that is injection of generics. Something like this `@Requires QueriesHandler<QueriesHandlerCommon.Query> queryHandler` will result in a random `QueriesHandler` being injected, because the type parameter cannot be read at runtime.

The solution is to introduce empty interfaces that extend the generic interface, like in the example I provided for `QueriesHandlerCommon`. I think this mostly affects implementations of `QueriesHandler`, but please check all usages of `@Requires` again for generics.

azahnen:
@bekakh I had a look at this, there are still 3 generic injections left.

azahnen (approved):
@cportele Looks good to me now, can you check if it works for you?

cportele (approved):
I tried it with a couple of configurations and various resources, no problems detected.

I will merge...

PR #406, opened 2021-03-17
Improve HTML and schema.org representation

### LINKED ISSUE #405: Improve HTML and schema.org representation
* Add `attribution` option in the service metadata so that this does not have to be included in the description and can be reported separately on the Landing Page (note: Common Part 1 will also add the attribution property).
* Add optional creator and publisher information to the metadata to improve the schema.org data (e.g., missing creator raises warnings in the Google validator) and so that creator and publisher can be properly represented.
* For datasets with multiple collections, report each collection as a sub-dataset (so that these can also be found in searches, often the "real" data is only described in the description of the feature type/collection). For cases with a single collection, that collection should not be reported as a sub-dataset due to the 1-to-1 relationship.
* Links with rel `enclosure` should be reported as distributions, too, in addition to the links with rel `items`.
* Support a `googleSiteVerification` configuration parameter (FOUNDATION module) that can be set in a default configuration like `apiCatalogLabel` to include a meta tag in the HTML page of the API catalog page.
* Search engine crawlers do not send accept headers, so we should ensure that the fallback for googlebot etc is HTML and not JSON.
* Provide breadcrumbs also as schema.org markup.

Internal cleanup:

* Computation of spatial/temporal extents for the Views is done multiple times and can be centralized.
* Rename `bbox2` field in View instances / mustache files to `bbox`.
* Fix double slashes in some `___static___` urls.

cportele:
Closes #405 

https://demo.ldproxy.net/ is currently using this branch.

The PR is not urgent.


azahnen (approved):
LGTM

PR #410, opened 2021-03-28
Temporal filtering improvements

### LINKED ISSUE #409: Temporal filter improvements
* Support dates before 1970-01-01 (allow negative values for milliseconds since epoch).
* If the `datetime` parameter uses dates, not timestamps, an interval that starts on the first second of the start date and ends at the last second of the end date should be used.
* ISO 8601 intervals include both the start and end instant. PostgreSQL intervals are exclusive of the end instant, so we have to add one second to each end instant before creating the SQL interval.
* Support intervals in `AFTER` and `BEFORE`.

cportele:
closes #409 

also requires merging https://github.com/interactive-instruments/xtraplatform-spatial/pull/52


azahnen (approved):

cportele:
replaced by #412

Issue #395, opened 2021-03-08
Update Filtering/CQL to Part 3, 1.0.0-draft.2


cportele:
Part 3, 1.0.0-draft.2 (in public review): https://portal.ogc.org/files/?artifact_id=96288&version=2

* [x] The BNF has been updated in general and we should sync with the current grammar. Updated BNF: https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/cql/standard/schema/cql.bnf. Important changes in predicates:
  * LIKE operator has additional options and by default comparisons are case insensitive
  * Array expressions have been added: https://docs.ogc.org/DRAFTS/19-079r1.html#arrays
  * Support double quoted identifiers (`doubleQuote`)
  * A boolean literal has been added
  * axis order of coordinates in ENVELOPE has changed (to align with bbox parameter)
  * Note: We won't implement any other parts of CQL that were excluded in the initial implementation
* [x] Keywords in the BNF grammar are case-insensitive. I.e., "and", "AND", "aNd" etc. all match "AND". This is a general rule for ABNF grammars. Currently, they the implementation is case sensitive. 
* [x] A query parameter `filter-crs` must be supported: https://docs.ogc.org/DRAFTS/19-079r1.html#filter-filter-crs

azahnen:
### Implementation tasks:

- **case insensitive keywords** 
  - [ ] define keywords in `CqlLexer.g4` as described [here](https://github.com/antlr/antlr4/blob/master/doc/case-insensitive-lexing.md#case-insensitive-grammars)
- **filter-crs**
  - [x] our CQL text parser already accepts a CRS parameter, currently it is always set to CRS84
  - [ ] add `read` method with additional CRS parameter in `Cql`/`CqlImpl`, pass to `CqlTextParser.parse`
  - [ ] add `QueryParameterFilterCrs`, read and pass to `CQL.read` in `FeatureQueryImpl`
- **double quoted identifiers**
  - [ ] adjust grammar in CqlLexer.g4`  
- **axis order of coordinates in ENVELOPE has changed**
  - [ ] adjust grammar in CqlParser.g4`  
- **LIKE operator has additional options**
  - [ ] adjust grammar in CqlParser.g4` and domain model
- **Array expressions** TBD
- **general**
[...]

PR #423, opened 2021-04-20
filter-crs query parameter

bekakh:
Add support for the `filter-crs` query parameter, which allows to encode encode geometric expressions from CQL filters in one of the supported CRSs. Addresses https://github.com/interactive-instruments/ldproxy/issues/395.


cportele (approved):
Open issues are documented in #432

PR #430, opened 2021-04-22
fix tile seeding

### LINKED ISSUE #429: tile seeding fails if some type has no geometry
Tile seeding will try to generate tiles even if the collection/type has no geometry and fails with `Filter is invalid. Unknown property: PROPERTY_NOT_AVAILABLE`.

azahnen:
Closes #429.

@cportele Can you check if this makes sense and if there are more places where this would be needed?


cportele (approved):
The changes look fine. I also tried it with collections without geometries.

Unrelated to this change to ldproxy, but one of the changes to xtraplatform(-spatial) today is now writing lots of the following entries in the log when the service starts:

`DEBUG [2021-04-22 20:37:08,184]                          - null: null -> LOADING  [d.i.x.s.d.entities.AbstractPersistentEntity] [startup]`

These should be either fixed or suppressed.

azahnen:
@cportele Sorry, I will remove the messages.

Issue #425, opened 2021-04-21
Some derived properties are written in the JSON output


cportele:
In some cases properties are derived, e.g., for the HTML mustache output, but they are not marked as `JsonIgnore` and are written in the JSON representations. The `JsonIgnore` annotation needs to be added.

The additional JSON members are not harmful as clients will ignore them, but they are redundant and unnecessary.

PR #433, opened 2021-04-28
update Value and JsonIgnore annotations

cportele:
- remove unnecessary `@Value` annotations from methods with parameters
- change `Derived` annotations to `Lazy` for methods on configuration classes that are only accessed in some circumstances
- add `JsonIgnore` for JSON API output for fields derived for the HTML output or for internal use
- cleanup code


cportele:
closes #425

PR #448, opened 2021-05-13
Dropwizard 2.0

### LINKED ISSUE #427: upgrade dropwizard from 1.3 to 2.0
1.3.29 is the last bugfix release for 1.3, so we need to migrate to 2.0.

According to the [Upgrade Notes](https://www.dropwizard.io/en/release-2.0.x/manual/upgrade-notes/upgrade-notes-2_0_x.html) I would not expect any larger issues. Besides thorough testing it comes down to:

- check if our custom JAX-RS context injections with HK2 still work
- migrate any remaining JAX-RS context field injections to method injections

azahnen:
Closes #427.

I migrated the recommended patterns, fixed all compile and runtime errors and tested some APIs. There don't seem to be any obvious issues.

I had to disable some tests in `SchemaValidatorSpec` which are no longer throwing NPEs. I would suspect that the default behaviour when parsing empty strings changed for `ObjectMapper`. Not sure how useful these tests were in the first place.


cportele (approved):
Looks good to me. I have updated the tests. I will not merge, because the dropwizard branch in xp is just a branch and not yet a PR.

PR #464, opened 2021-06-23
catch exceptions due to invalid headers during content negotiation

### LINKED ISSUE #463: broken HTTP header = silent failure
An invalid HTTP header (in this case `Accept-encoding: ,gzip` for whatever reason sometimes sent by uptimerobot.com) resulted in a ParseException during content negotiation returning a 500 response, but without logging anything. We should log a warning and pick any content type/language.

cportele:
closes #463


azahnen (approved):
LGTM

Issue #469, opened 2021-07-08
add admin task to purge files from a tile cache


cportele:
Options:

* the API (mandatory)
* a tile matrix set to select the tile sets to purge (default: purge tile sets for all tile matrix sets) 
* a collection to select the tile sets to purge (default: purge tile sets for all collections as well as the dataset tile sets); the dataset tile sets will always be purged, too
* a WGS84 bbox to select the tiles per zoom level in each tile set (default: no spatial filter, all tiles per zoom level)

The TileCache class already includes methods for purging (via #470), but those methods need testing (and tests).

azahnen:
I implemented an admin task that just passes the (validated) parameters to `TileCache.deleteTiles`. Example:

```
curl -X POST "http://localhost:7081/tasks/purge-tile-cache?api=vineyards&collection=vineyards&tileMatrixSet=WebMercatorQuad&bbox=8,49,9,50"

```

cportele:
The issue will be closed after more testing.

azahnen:
I added support in `xtractl`: https://github.com/interactive-instruments/xtraplatform-cli/releases/tag/v1.1.0

cportele:
All tasks have been implemented and tested

Issue #467, opened 2021-07-02
support mbtiles as an additional option for the tile cache


cportele:
Addition of a new option in the TILES configuration:

* `cache: FILES` (default)
* `cache: MBTILES`

For FILES everything is as before.

The simplest approach for MBTILES would be:

For each `cache/tiles/{api}/{collectionId}` directory (for the dataset tiles "__all__" is used as the collectionId) a mbtiles file is created with the name of the tile matrix set. The mbtiles structures are quite simple, so we should manage them directly via SQL statements. 

The non-cachable tiles are continued to be created in a tmp directory in the file system and deleted after a short time.

cportele:
Changes were moved from PR #450 to #470

Issue #302, opened 2020-10-29
optimize cache for empty tiles


cportele:
An empty vector tile in the cache is a file with zero bytes, but these still consume disk space for the file system index. For sparse data and larger zoom levels, there may be many empty tiles. If these would be registered in the cache in a different way and not stored in the file system, this could save significant disk space in some cases.

cportele:
Changes were moved from PR #450 to #470

PR #470, opened 2021-07-08
Tile cache

cportele:
Changes:
- support MBTiles as tile cache (one file per tile set) (#467, already closed)
- in MBTiles caches, empty MVT tiles are represented by a single BLOB (#302, already closed)
- add capability to purge tiles from the tile cache (by collection, bbox and/or tile matrix set) in preparation for future admin task (see #469)


azahnen (approved):
LGTM. I ran some tests with both `FILES` and `MBTILES`, works as expected.

cportele:
Yes, I will resolve and then merge...

PR #478, opened 2021-07-20
support caching headers / conditional GET requests

### LINKED ISSUE #475: Support caching headers / conditional GET requests
Support caching headers and conditional GET requests in all current endpoints. 

Response headers:

* `Last-Modified`
* `ETag`
* `Cache-Control`
* `Expires`

Request headers:

* `If-Match`
* `If-None-Match`
* `If-Modified-Since`
* `If-Unmodified-Since`

The request headers can be evaluated using the JAX RS `evaluatePreconditions()` method on `Request`.

The response headers have to be added in the responses. The rules partly depend on the resource type (see below), but there need to be some configuration options and general rules to provide useful header values.

Add configuration options for the service configuration (default values, `defaultCaching`) and for each module with GET endpoints (`caching`). The configuration object can have the following optional values:

* `lastModified`: A UTC timestamp that is used in some cases to set a `Last-Modified` header. This can be useful, in cases where a server is stable, but the built-in default mechanism does not work.
* `expires`: A UTC timestamp that can be use to set an `Expires` header. This can be useful, if it is known, that the resources will change at a certain time in the future, e.g. for period data updates. If this configuration option is not set, no `Expires` header will be written.
* `cacheControl`: The text value of the `Cache-Control` header. If this configuration option is not set, no `Cache-Control` header will be written. If is recommended to set this header with a reasonable value, which will depend on the stability of the API.
* `cacheControlItems`: This option can be used to override the `cacheControl` value in modules that include a collection and and items resource (e.g. Features, Collections, TileSets, TileMatrixSets, Styles, Resources). In this case, this can be used to set a different value for the items. This can be used, for example, in cases where the collection is dynamic, but once an item has been added it is stable - like a feature collection where new features are added continuously, but once a feature has been added, it is stable.

`Cache-Control` and `Expires` header will be written, if configured. 

For `Last-Modified` the following variants are needed, depending on how the resource is managed:

* a file: determine the information from the file system
* from files: determine the information from a relevant files in the file system, use the latest value
* from configuration: use a `lastModified` configuration or set to the value of the last-modified timestamp in the service configuration - and in some cases the provider configuration
* now: use a `lastModified` configuration or set to now

For `ETag` the following variants are needed:

* byte array: compute from a byte array or a file that will be returned
* POJOs: compute from a PageRepresentation or sub-type POJO that will be returned, a weak ETag
* last-modified: compute from the last-modified date, a weak ETag

The computation needs to be stable across restarts of the JVM, should minimize collisions and be fast.

In order to compute stable values for POJOs, they need a mechanism to compute a stable string representation, which needs to be added to each class. 

|  Resource | Last-Modified | ETag | Remarks | 
|---|---|---|---|
| Landing Page | from configuration | POJO |  | 
| Conformance Declaration | from configuration | POJO |  | 
| API Definition | from configuration | - | no simple way to compute a hash |
| Collections | from configuration | POJO |  |
| Collection | from configuration | POJO |  |
| Features | from configuration | - | the last-modified timestamp in the provider configuration is considered, too; no simple way to compute a hash |
| Feature | from configuration | last-modified | no simple way to compute a hash, computation of a hash for a feature should be a convenience function of the provider, this will be important for conditional PUT/PATCH requests in the transaction module |
| Queryables | from configuration | POJO |  |
| Schema | from configuration | POJO |  |
| Tile Sets | from configuration | POJO |  |
| Tile Set | from configuration | POJO |  |
| Tile | now | - | for single-layer tiles from a feature provider that are computed with a feature transformer; no simple way to compute a hash |
| Tile | now | from byte array | for multi-layer tiles from a feature provider that are computed from the single-layer tiles |
| Tile | from Mbtiles file | from byte array | for tiles from an Mbtiles provider |
| Tile | from tile file | from byte array | for tiles from a file cache |
| Tile | now | from byte array | for empty tiles |
| Tile Matrix Sets | from configuration | POJO |  |
| Tile Matrix Set | from configuration | POJO |  |
| Styles | from files | POJO |  |
| Style | from file | POJO |  |
| Style Metadata | from files | POJO |  |
| Resources | from files | POJO |  |
| Resource | from file | from byte array |  |

Note: This issue is only about GET (and HEAD) endpoints. OPTIONS, PUT, DELETE, PATCH also need to support the headers.

cportele:
closes #475 

The code has been deployed in the [t17-APIs](https://t17.ldproxy.net).

Guava is used for the ETag hashing with murmur3_128 as the hash function based on this [comparison](http://goo.gl/jS7HH) linked from the [Javadoc](https://guava.dev/releases/snapshot/api/docs/com/google/common/hash/Hashing.html).

### DIFF
--- a/ogcapi-draft/ogcapi-collections-queryables/src/main/java/de/ii/ldproxy/ogcapi/collections/queryables/app/EndpointQueryables.java
+++ b/ogcapi-draft/ogcapi-collections-queryables/src/main/java/de/ii/ldproxy/ogcapi/collections/queryables/app/EndpointQueryables.java
@@ -121,13 +121,9 @@ public class EndpointQueryables extends EndpointSubCollection /* implements Conf
                              @Context UriInfo uriInfo,
                              @PathParam("collectionId") String collectionId) {
 
-        boolean includeLinkHeader = api.getData().getExtension(FoundationConfiguration.class)
-                .map(FoundationConfiguration::getIncludeLinkHeader)
-                .orElse(false);
-
         QueryablesQueriesHandlerImpl.QueryInputQueryables queryInput = new ImmutableQueryInputQueryables.Builder()
+                .from(getGenericQueryInput(api.getData()))
                 .collectionId(collectionId)
-                .includeLinkHeader(includeLinkHeader)
                 .build();
 
         return queryHandler.handle(QueryablesQueriesHandlerImpl.Query.QUERYABLES, queryInput, requestContext);
--- a/ogcapi-draft/ogcapi-collections-queryables/src/main/java/de/ii/ldproxy/ogcapi/collections/queryables/app/QueryablesQueriesHandlerImpl.java
+++ b/ogcapi-draft/ogcapi-collections-queryables/src/main/java/de/ii/ldproxy/ogcapi/collections/queryables/app/QueryablesQueriesHandlerImpl.java
@@ -20,6 +20,7 @@ import de.ii.ldproxy.ogcapi.domain.QueryHandler;
 import de.ii.ldproxy.ogcapi.domain.QueryIdentifier;
 import de.ii.ldproxy.ogcapi.domain.QueryInput;
 import de.ii.ldproxy.ogcapi.features.core.domain.FeaturesCoreProviders;
+import de.ii.ldproxy.ogcapi.features.geojson.domain.JsonSchema;
 import de.ii.ldproxy.ogcapi.features.geojson.domain.JsonSchemaObject;
 import de.ii.ldproxy.ogcapi.features.core.domain.SchemaGeneratorFeature;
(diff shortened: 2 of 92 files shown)


azahnen (approved):
The implementation looks good to me.The concept described in #475 looks solid, but I only skimmed through it. I guess there was no way to reuse the toString and/or hashCode methods autogenerated by immutables? It seems like quite the overhead to provide that manually for every new class.

The changes around JsonSchema overlap with my current work, so I should have reviewed this earlier :-) Let's merge this ASAP if there are no general concerns, so that I can try to merge the changes.

cportele:
The main issue with hashCode() is that the hashCode() will be different with a new execution of the app. toString() has similar issues as a result, for example, if a POJO contains a set - the results differs also between two copies. I only added all the manual code generation after I figured out why the simpler approach didn't work and after some research I was not able to find a simpler solution.

Feel free to merge...

PR #482, opened 2021-07-25
support 'now' in datetime parameter and in CQL filters

### LINKED ISSUE #481: support "now" as a temporal value in datetime/filter
Extend the value range of the `datetime` parameter to also support "now" as an instant and convert it to the current timestamp in seconds. Also allow "now" for the start or end instant in an interval. For example:

* `datetime=now`
* `datetime=now/..`

Also extend CQL to offer the same capabilities. For a temporal instant use "now()" to avoid parsing conflicts with a property name.

This is an extension to the OGC standards.

cportele:
closes #481

See also https://github.com/interactive-instruments/xtraplatform-spatial/pull/83 for the changes in xtraplatform-spatial. The main changes are in that PR to extend the CQL grammar and to handle 'now' as part of the TemporalLiteral class.

### DIFF
--- a/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ldproxy/ogcapi/features/core/app/FeaturesQueryImpl.java
+++ b/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ldproxy/ogcapi/features/core/app/FeaturesQueryImpl.java
@@ -352,7 +352,8 @@ public class FeaturesQueryImpl implements FeaturesQuery {
     private Optional<CqlPredicate> timeToCql(String timeField, String timeValue) {
         // valid values: timestamp or time interval;
         // this includes open intervals indicated by ".." (see ISO 8601-2);
-        // accept also unknown ("") with the same interpretation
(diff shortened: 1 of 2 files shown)


azahnen (changes requested):
In filters it is only working with `now()` for me. So `now() AFTER unfzeit` works, but `now AFTER unfzeit` returns: 

```
The parameter 'filter' is invalid. Unknown or forbidden properties used: now.
```

I guess the parentheses would have to be made optional in the grammar:

https://github.com/interactive-instruments/xtraplatform-spatial/pull/83/files#diff-97fb390aae582b3662058313bcde28d34edf8f5702773a070586ed4f93643ff7

cportele:
This use of a function in CQL in case of an instant is intentional, because otherwise it is not possible to disambiguate `now` as a special timestamp and a property reference `now`. I actually first had the parentheses optional (or even not there at all) and that lead to parsing problems complaining about not queryable `now`, IIRC.

azahnen (approved):
Ok, then let's merge this.

Issue #480, opened 2021-07-22
Support date/datetime values as sort keys and in CQL comparison predicates


cportele:
Note: Only for SQL feature providers.

Should be implemented together with #335.

azahnen:
I pushed a commit in interactive-instruments/xtraplatform-spatial#84 to support date/datetime as secondary sort keys. Note that using them as primary sort keys in the provider will lead to unexpected results as date/datetime values will not be unique in most cases.

cportele:
@azahnen - I have tested the sorting with the NOTAM API from T17 ([link to the deployed data](https://t17.ldproxy.net/fns/collections/notam/items)), by adding

```yaml
    - buildingBlock: SORTING
      enabled: true
      sortables:
      - issued
```
`sortby=issued` works, but `sortby=-issued` results in no features (but numberMatched is 10).

azahnen:
Is it possible that `issued` contains null values? I tested with geoval observations and `-resultTime.start` worked as expected. I think the code does not expect null values when creating the min/max filters for sort keys, so I would have to fix that.

cportele:
No, there is no `null` value ([query link](https://t17.ldproxy.net/fns/collections/notam/items?filter=issued+is+null&f=json)). 

But, in general, sortby should also work with `null` values, not just for instants.

azahnen:
I agree, but that might have explained it. I will test with t17 tomorrow.

azahnen:
I pushed a fix for the datetime issue, which occurred due to the timezone being ignored. Ascending did not work correctly either, but since the page showed 9 features, it wasn't so obvious.

That being said, after some more tests I came to the conclusion that the current implementation does not support non-unique sort keys. I tried to find an easy fix, but it will require some deeper changes to the query generation, so I created a separate issue #488.

azahnen:
@cportele This should work as expected in `xtraplatform-spatial` version `5.0.0-sql-queries-SNAPSHOT`.

cportele:
@azahnen - I tried with 'gebaeude' and am getting strange results with `sortby=name` and `sortby=-measuredHeight`. I have not yet tried to debug why that happens.

PR #483, opened 2021-07-27
redesign feature transformation engine

### LINKED ISSUE #165: refactor json schema generation
Depends on #162.

Two topics should be addressed:

### Transformations

Schema transformations, e.g. flattening of objects, should be the responsibility of the provider. Such transformations should be defined either in the provider configuration or for a specific output format in the service configuration. A consumer should then be able to get the transformed schema from the provider.

### Encoding

Schemas might be encoded in different output formats, just like features. Therefore we should implement a similar mechanism to register encodings with a mime type.

### LINKED ISSUE #190: refactor feature transformers
In #162 a new way to process features was implemented. The idea is that all transformations are implemented in a generic, output format agnostic way and are handled by the provider. 

That would include schema transformations like flattening of objects, value transformations like formatting and codelists and also all coordinate transformations. This will make the code less redundant and allows for better testability of the isolated transformations.

Another part of that is a generic way to build up complex objects from the seperate feature tokens, like we currently do in FeatureTransformerHtmlComplexObjects.

So the goal of this issue would be to make use of the new features and reimplement the current FeatureTransformers as much slimmer FeatureEncoders.

### LINKED ISSUE #323: Queryables and datetime/bbox
Currently the order of the temporal queryables in the service configuration is relevant, if multiple queryables are specified. The first is the start and the second the end of an interval when the `datetime` parameter is used. 

This should be made explicit in the provider configuration using `INSTANT`, `START` and `END` as roles for the temporal properties. 

In case of multiple spatial properties, we also need a role like `PRIMARY` to identify the property that should be used for `bbox`.

### LINKED ISSUE #335: Schema: Distinguish Date and DateTime properties
Date and DateTime properties are both represented as DATETIME, so we cannot declare to the user / in the queryables / in the schema, whether a temporal property is a DateTime or a Date.

### LINKED ISSUE #472: timestamps must conform to RFC 3339
Temporal properties must include an explicit time zone in the timestamp, but currently timestamps are provided without a timezone from the the feature transformer.

The timezone should be fixed to UTC ('Z') for timestamps returned from the API - at least as a default.

Also, the character between the date and the time has to be 'T', not a blank.

azahnen:
Closes #190.
Closes #335.
Closes #472.
Addresses #480.
Closes #165.
Closes #323.

### Todo
- [x] Adapt Schema module to support flatten transformation. See #165.
- [x] If the flatten transformation uses another separator than ".", the "." is still used in the Queryables and the Projections. This is confusing and in this case, the other separator should be used for the property names, too.
- [x] wfs/gml decoder
- [x] tiles encoder
- [x] projections
- [ ] documentation 

### Follow-up
- observation processing encoder (community module)
- flatgeobuf encoder (community module)

#### GeoJSON

`nestedObjectStrategy`, `multiplicityStrategy` and `separator` are deprecated. These are superseded by the generic `flatten` transformation. If necessary the transformation is automatically added for backwards compatibility. 

So for example this

```yml
- buildingBlock: GEO_JSON
  nestedObjectStrategy: FLATTEN
  multiplicityStrategy: SUFFIX
  separator: '.'
```

would be transformed to this

```yml
- buildingBlock: GEO_JSON
  transformations:
    '*':
      flatten: '.'
```

#### HTML

`layout` is deprecated. It is superseded by `mapPosition` and the generic `flatten` transformation. If necessary `mapPosition` is overriden and the transformation is automatically added for backwards compatibility. 

So for example this

```yml
- buildingBlock: FEATURES_HTML
  layout: CLASSIC
```

would be transformed to this

```yml
- buildingBlock: FEATURES_HTML
  mapPosition: RIGHT
  transformations:
    '*':
      flatten: '.'
```

`mapPosition` can be `TOP`, `RIGHT` or `AUTO`. `AUTO` is the default, it chooses `TOP` when any nested objects are found and `RIGHT` otherwise (only when `layout` is not set, otherwise it will be overriden). 

#### Provider

Support for type `DATE` (#335) and support for `nativeTimeZone` (#472).

Transformations can now be defined in the provider as well. Transformations supporting the wildcard `*` (currently only `flatten`) can be defined in the feature, all other transformations can be defined in the properties, for example:

```yml
types:
  umweltzone:
    type: OBJECT
    sourcePath: /am_managementrestrictionorregulationzone{filter=id_theme=1}
    transformations:
      flatten: '/'
    properties:
      inspireId:
        sourcePath: inspireid
        type: STRING
        transformations:
          stringFormat: https://daten.nrw.de/id/umwelt/luftreinhalteplan/{{value}}

```


ogcapi-stable/ogcapi-features-html/src/main/java/de/ii/ldproxy/ogcapi/features/html/app/FeatureEncoderHtml.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+
+  @Override
+  public void onFeature(FeatureHtml feature) {
+    transformationContext.featuresHtmlConfiguration().getFeatureTitleTemplate().ifPresent(label -> transformationContext.collection().name = label);
+
+    //TODO
+    if (!feature.hasObjects()) {
+      feature.getProperties().forEach(propertyHtml -> propertyHtml.name(propertyHtml.getSchema().map(
```
cportele, line 141: I think this should be

```
feature.getProperties().forEach(propertyHtml -> propertyHtml.name(propertyHtml.getSchema().flatMap(
          FeatureSchema::getLabel).orElseGet(propertyHtml::getName)).propertyPath(ImmutableList.of(propertyHtml.getName())));
```

cportele:
I will record issues that I have identified in tests in this comment:
* [ ] In the HTML output, the labels are not used. See the code comment with a change proposal.
* [ ] When starting the geoinfodok API, the following provider error is reported:

  ```
  Feature provider with id 'geoinfodok' could not be started: not a valid relation path: record{filter=catalog_fk=1}/[id=record{filter=catalog_fk=1}_fk]record{filter=catalog_fk=1}_association
  ```

  The second and third filter expressions should not be there. The correct path according to the provider configuration would be `record{filter=catalog_fk=1}/[id=record_fk]record_association`.

ogcapi-stable/ogcapi-features-html/src/main/java/de/ii/ldproxy/ogcapi/features/html/app/FeatureEncoderHtml.java
azahnen: This should have catched only the classic/flatten case, which does not use labels. But it was also applied to schemas that were flat in the first case. I implemented a better solution.

azahnen:
I checked the items that should be fixed.

The MVT encoder is ready for testing. It is currently 50% slower, I will fix that when implementing the community encoders.

cportele:
I did some tests with the tile encoder. Problems:

* [ ] With the vineyards data an exception is thrown, see below for the stacktrace (multi-polygon geometry).
* [ ] With the daraa data, only layers with point geometries are added to the tiles. (I will see, if I can find out more information.)

```
ERROR [2021-08-02 17:45:25,971]                vineyards - Server Error with ID c32ff0a7ca3b17f6: Points of LinearRing do not form a closed linestring [9f7f5b71-2df9-46ed-a3f0-e5ca7b2af2a0] [d.i.l.ogcapi.infra.rest.ExceptionMapper] [dw-330]  
DEBUG [2021-08-02 17:45:25,973]                vineyards - Stacktrace: [9f7f5b71-2df9-46ed-a3f0-e5ca7b2af2a0] [d.i.l.ogcapi.infra.rest.ExceptionMapper] [dw-330]  
! java.lang.IllegalArgumentException: Points of LinearRing do not form a closed linestring
! at org.locationtech.jts.geom.LinearRing.validateConstruction(LinearRing.java:93)
! at org.locationtech.jts.geom.LinearRing.<init>(LinearRing.java:88)
! at org.locationtech.jts.geom.GeometryFactory.createLinearRing(GeometryFactory.java:382)
! at org.locationtech.jts.geom.GeometryFactory.createLinearRing(GeometryFactory.java:369)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.getJtsCoordinateArrayAsGeometry(PropertyMVT.java:127)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.getJtsLinearRing(PropertyMVT.java:80)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.lambda$getJtsPolygon$6(PropertyMVT.java:93)
! at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
! at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
! at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
! at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
! at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:550)
! at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
! at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:517)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.getJtsNestedGeometryArray(PropertyMVT.java:120)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.getJtsPolygon(PropertyMVT.java:93)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.lambda$getJtsMultiPolygon$8(PropertyMVT.java:102)
! at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
! at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
! at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
! at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
! at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:550)
! at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
! at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:517)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.getJtsNestedGeometryArray(PropertyMVT.java:120)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.getJtsGeometryArray(PropertyMVT.java:112)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.getJtsMultiPolygon(PropertyMVT.java:102)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.lambda$getJtsGeometry$1(PropertyMVT.java:57)
! at java.base/java.util.Optional.flatMap(Optional.java:294)
! at de.ii.ldproxy.ogcapi.tiles.app.PropertyMVT.getJtsGeometry(PropertyMVT.java:44)
! at de.ii.ldproxy.ogcapi.tiles.app.FeatureMVT.lambda$getJtsGeometry$4(FeatureMVT.java:95)
! at java.base/java.util.Optional.flatMap(Optional.java:294)
! at de.ii.ldproxy.ogcapi.tiles.app.FeatureMVT.getJtsGeometry(FeatureMVT.java:95)
! at de.ii.ldproxy.ogcapi.tiles.app.FeatureEncoderMVT.onFeature(FeatureEncoderMVT.java:134)
```

azahnen:
That is strange, I tested with vineyards, daraa and strassen from the demo configuration and got the exact same results as with master. I will have another look tomorrow.

cportele:
OK, this is weird. Then let me check my set up first to see that the correct components are used...

cportele:
The Vineyards exception does not occur when I execute in composite. It occurs when execute in ldproxy with "-Psnapshot", but this uses outdated code since the last build in the CI have failed, which may be the reason for the behaviour. So I think we can ignore that one.

However, in composite I also get an exception when requesting tile `9/172/264`. This happens because there is no feature in the tile. I get the same error, if I do a feature query with `bbox=0,0,1,1`.

```
DEBUG [2021-08-03 12:32:20,017]                vineyards - Processing request: GET /tiles/WebMercatorQuad/9/172/264 [a556e2ef-0982-46bf-8537-23ea6a88ff26] [d.i.x.services.app.ServicesEndpoint] [dw-93]  
...
TRACE [2021-08-03 12:32:20,039]                vineyards - SUB FILTER: (A.ogc_fid >= 'null' AND A.ogc_fid <= 'null') AND (A.ogc_fid IN (SELECT AA.ogc_fid FROM vineyards AA  WHERE ST_Intersects(AA.wkb_geometry, ST_GeomFromText('POLYGON((259603.80628787805 5576251.150272974,309678.81308622775 5574216.482906233,311469.04023684363 5623922.971657132,261865.84599338114 5625951.4687801255,259603.80628787805 5576251.150272974))',25832)))) [a556e2ef-0982-46bf-8537-23ea6a88ff26] [d.i.x.f.p.s.a.FeatureStoreQueryGeneratorSql] [stream.vineyards-8]  
ERROR [2021-08-03 12:32:20,043]                vineyards - Server Error with ID e7bc2b8d63db9ad4: Unexpected SQL query error: 
  ERROR: invalid input syntax for integer: "null"
  Position: 271 [a556e2ef-0982-46bf-8537-23ea6a88ff26] [d.i.l.ogcapi.infra.rest.ExceptionMapper] [dw-93]  
```

azahnen:
The issue was introduced by the sort key changes, I pushed a fix.

cportele:
Yes, that fixed the error. I still only get points in Daraa, I will have a look for the cause.

cportele:
With Daraa, the MultiPolygon or MultiLineString geometries are converted to empty although the nested geometry properties are there, down to the coordinates. I do not really understand yet why, because vineyards has MultiPolygons, too, and there the geometry is converted to proper JTS geometries. I will continue tomorrow...

azahnen:
I pushed a fix for the MultiPoint as well as for the daraa issue. The problem with daraa was that the configuration contains the wrong geometry types (without MULTI_) which were used in one place instead of the actual type. The old code adjusted the schema that was passed down to the transformer with the actual type, I replicated that in the new code. An open todo in the old code was if we should at least log a warning or what else the right way to handle this case would be.

cportele:
Yes - the changes fixed these issues. I think a warning in the log would be appropriate, but on the other hand, if I request thousands of features that is a bit overwhelming. Maybe a strict errors check during startup that can catch the systematic configuration issues during the configuration development and then a log entry for individual cases would be ok?

(FWIW, I had noticed the MULTI issue while debugging and tried to change it for one feature type, but that had no effect. But it works with the updated code. I will update the Daraa configuration so that the geometry types match.)

cportele:
I have compared seeding times for the demo APIs with older notes and they are in the same range.

cportele:
I have also tested transformations in the provider schema by moving all Features Core transformations to the provider schema. I did not find any issues expect that '*' was reported as an invalid property key during startup validation. I have pushed a fix.

cportele:
I have made tests with the updated code. I have added two fixes as a result. Projections are working now.

I identified a new issue that has been introduced in the xtraplatform-spatial changes: while I can also use nested properties in projections (e.g. "strasse.klasse"), they are now rejected by the SQL encoder.

azahnen:
@cportele 
- Please confirm that tiles are working as expected.
- Nested properties in filters should work again (`strasse.klasse`).
- Nested properties for `gebaeude` should work now.
- The mapping filter for `gebaeude` works if you comment out the blocks for `consistsOfBuildingPart` and `parent`. Self joins are currently not supported, there are multiple places in the code where it is assumed that a table only occurrs once in an absolute source path.

cportele:
Thanks @azahnen. I will check Tiles next week, gebaeude looks much better now, I will have a closer look also next week. We will have to support self joins, but we can do that after next week.

azahnen:
@cportele 
- I pushed the schema generation refactoring.
- I made a short test with multiple geometries, it is not working yet. We would have to adjust `GeoJsonWriterGeometries` to handle primary and secondary geometries correctly.

cportele:
Thanks @azahnen. I will test all encoders again and try to fix issues myself, if I find any.

cportele:
@azahnen - The last changes to FeatureSchema in https://github.com/interactive-instruments/xtraplatform-spatial/commit/a4ec7e987f58ffe5eaa011ffcaaaa0963614d25e are incomplete and have broken the GeoJSON encoder. 

The `@Value.Check` method `default FeatureSchema primaryGeometry()` incorrectly adds the role PRIMARY_GEOMETRY to geometries of nested objects (like the address or the surfaces in the gebaeude provider). These properties are not primary geometries and should not be treated as such. This also applies to `primaryInstant()` and `primaryInterval()`. In the geometry case a result is that the encoder gets into an incorrect state (buffering, creating `geometry` and `properties` member in the nested object, etc.) and stops with an exception.

The role of and the need for these new check functions is unclear to me, so I have not tried to create a fix.

azahnen:
@cportele Please try again.

The methods set the primary roles if they are not set explicitely in the mapping, so that the isPrimary* methods work as expected. They should run only on the feature level, but that did not work because parentPath is always emtpy in this context.

cportele:
@azahnen - Yes, that fixed the issue, thanks.

Issue #30, opened 2016-01-11
Add number of features on dataset pages?


cportele:
... in addition to name, description, keywords and extent. 

Only do this, if numberMatched has a value, i.e. this does not require an additional WFS request.

PR #487, opened 2021-07-31
Feature changes

cportele:
Changes in this pull request:

* Extent information for a collection is dynamic and has to be updated, if the underlying data in the collection changes. Move the immutable extent information from OgcApiDataV2 to a new component (CollectionDynamicMetadataRegistry). The extent information as specified in the configuration is still part of OgcApiDataV2.
* In addition to the spatial and temporal extent, define metadata elements for the last modification (for caching headers) and the item count in the collection (see #30).
* Move the computation of the bboxes/intervals from the hydration to the startup and populate the metadata entries.
* Add a capability to process notifications about feature changes and update the metadata registry.
* Add a listener for PostgreSQL notifications. The channel name for the notification must be the collectionId. The value of the notification is a CSV record with the following entries:
  * `INSERT`, `UPDATE` or `DELETE`
  * the table name (currently not used)
  * the featureId
  * two timestamps (a RFC 3339 date-time, `NULL` or empty)
  * four WGS84 bbox values (number or empty, standard order: west, south, east, north)

If the collection or change has no temporal information, the two timestamps will be empty. Otherwise, if the collection has a single temporal queryable the first timestamp will be set and the second will be empty and if the collection has two temporal queryables both timestamps will be set representing start and end (`NULL` indicates an open start or end).

If the collection or change has no temporal information, the four bbox numbers will be empty. Otherwise all four will be set.

Here is an example of a PostgreSQL trigger function:

```
CREATE OR REPLACE FUNCTION public.notam_trigger()
 RETURNS trigger
 LANGUAGE plpgsql
AS $function$
   DECLARE
      bbox box2d;
   BEGIN
        IF (TG_OP = 'DELETE') THEN
            bbox = box2d(ST_Transform(OLD.geom,4326));
            PERFORM pg_notify(TG_TABLE_NAME, concat_ws(',', TG_OP, TG_TABLE_NAME, OLD._id, COALESCE(to_char(OLD.valid_time_begin, 'YYYY-MM-DD"T"HH24:MI:SS"Z"'),'NULL'), COALESCE(to_char(OLD.valid_time_end, 'YYYY-MM-DD"T"HH24:MI:SS"Z"'),'NULL'), st_xmin(bbox), st_ymin(bbox), st_xmax(bbox), st_ymax(bbox)));
            RETURN OLD;
        ELSIF (TG_OP = 'INSERT') THEN
            bbox = box2d(ST_Transform(NEW.geom,4326));
            PERFORM pg_notify(TG_TABLE_NAME, concat_ws(',', TG_OP, TG_TABLE_NAME, NEW._id, COALESCE(to_char(NEW.valid_time_begin, 'YYYY-MM-DD"T"HH24:MI:SS"Z"'),'NULL'), COALESCE(to_char(NEW.valid_time_end, 'YYYY-MM-DD"T"HH24:MI:SS"Z"'),'NULL'), st_xmin(bbox), st_ymin(bbox), st_xmax(bbox), st_ymax(bbox)));
            RETURN NEW;
        END IF;
    END;
$function$
;

CREATE TRIGGER notam_update
AFTER INSERT OR UPDATE OR DELETE ON notam
    FOR EACH ROW EXECUTE PROCEDURE notam_trigger();
```

Still to do (create new follow-up issues):

* [ ] The metadata is currently populated during startup in EndpointFeatures. Should this be done somewhere else?
* [ ] Add a capability to periodically re-initialize the metadata from the feature data to account for lost notifications and/or because deletes cannot be not taken into account to update an extent.
* [ ] FeatureQueries.getFeatureCount() currently always returns 0, so for now the item count metadata element is not usable
* [ ] The PostgreSQL listener should be moved to the SQL feature provider.
* [ ] The SQL feature provider also caches extent information, the whole data flow needs some consolidation.
* [ ] Update documentation.


azahnen:
Replaced by #630.

PR #511, opened 2021-09-29
Weblibs

### LINKED ISSUE #495: Avoid external resources in HTML
Referencing external stylesheets (css) and scripts (js) in the HTML pages can be undesirable for multiple reasons:

* performance (network access)
* network issues (firewall or proxies blocking access)
* security concerns (access to third party resources)

These resources should be hosted by the application.

### LINKED ISSUE #496: Consolidate use of mapping libraries, use default style for all maps
Currently STYLES uses Mapbox GL JS, TILES uses OpenLayers (always with a neutral style), FEATURES uses Leaflet (again, with a neutral style, also to select the values of the bbox parameter in the Filter Editor) and COMMON/COLLECTIONS use Leaflet (to display the bounding box). 

Approach:

* Use of Mapbox GL JS (or MapLibre GL JS). Exception: Use of OpenLayers for TILES, if other tiling schemes shall be displayed besides WebMercatorQuad (configurable).
* In FEATURES_HTML as well as in TILES a style can be configured. If no style is configured, the defaultStyle from STYLES is used. If this is not configured, a simple default style is used. This should also be possible if STYLES in not enabled, then the styles would only be accessible internally.
* With the TILES the configured style can be used directly.
* Prerequisite for styles for use in FEATURES_HTML:
  * Two source entries: the vector tiles and the background map.
  * The source layers of the vector tiles correspond to the IDs of the collections.
  * Any properties used in the styles must be identical in the vector tiles and in GeoJSON.
  * 1..n layers are specified for each collection.
* The requirements should be checked to the extent possible.
* When using the styles in FEATURES_HTML, some changes have to be made automatically.

azahnen:
Closes #495.
Closes #496.

### Todo
- [ ] support styles for items in `FeatureCollectionView` (CP)
- [ ] support styles for tiles in `TileSetsView` (CP)
- [ ] filter editor (AZ)
- [ ] OpenLayers (AZ)
- [ ] layer toggler (AZ)

### Follow-Up
- Cesium


cportele:
@azahnen - Is something missing? There is `{{> style-common}}` is `html-start.mustache`, but it does not seem to be there.

cportele:
Forget the question... After another build I can see that the file got generated.

cportele:
The map seems to start with Germany and then fly to the data (this is a page reload):
 
reload

cportele:
@azahnen - I have pushed the following changes:

* support a style for features
* support a style for tilesets
* improvements for the derivation of styles
  * support tilesets without layer information (e.g. from mbtiles)
  * better filter for the geometry type
  * support multiple geometry types per collection
  * consider geometry type also for all layers
* make selection of the map client library an explicit configuration option
* prepare hash support
* set the initial view to (0,0) at zoom level 0

Open issues (also added to the initial comment):

* Add hash support (`hash: true` initializes maps at (0,0) and zoom level 10 for some reason, even, if the style has center/zoom information).
* I frequently get "Uncaught TypeError: this.stylesheet is undefined" errors. This error no longer appears after a page reload (or two). Reason is unclear.
* It would be good, if the initial view is set to the bbox of the dataset or collection.

PR #512, opened 2021-10-04
Integrate PROJ and CrsInfo

### LINKED ISSUE #398: Support NTv2 coordinate conversions
Where available, it should be possible to configure the use of NTv2 grid shift files for coordinate conversions.

### LINKED ISSUE #439: Use PROJ for CRS transformations instead of Geotools
An initial implementation that mostly works was done in #512. But intensive tests showed that PROJ-JNI is not as production-ready as we hoped:

- [x] shutdown due to out of memory exception in PROJ_JNI for certain CRSs, with an underlying file not found error in PROJ, most likely due to a missing grid file (also see https://github.com/interactive-instruments/ldproxy-project-management/issues/129#issuecomment-988712640 ff.)
- [x] memory leak when transformer/operation caching is disabled (see https://github.com/interactive-instruments/ldproxy-project-management/issues/129#issuecomment-988733272). 
- [x] `point outside of projection domain` errors with transformations results of `infinity`, where GeoTools returned at least some however nonsensical values, has to be handled in higher layers like tile generation
  - I think we can leave this like it is, it should not occur anymore when #594 is implemented. The error only occurs for really abnormal values. In the case of lika the affected coordinate is near Sumatra, and a point slightly northwards of that works as expected.
- [x] add configuration option for PROJ configuration & data directory, or at least allow to use the default behavior how PROJ locates the resource directories (https://proj.org/resource_files.html). Also make sure that ldproxy will not overwrite PROJ 
  - I added an option to `cfg.yml`:
    ```yml
    proj:
      location: /home/zahnen/proj
    ```
- [x] GeoTIFF support
- [ ] support RHEL7

Furthermore we identified some follow-up issues regarding the `xtraplatform-crs` api:

- [x] 3D transformers cannot handle 2D coordinates
- [x] cleanup api + implementation, e.g. `CoordinateTupleWithPrecision`, `CrsTransformerProj.isSourceMetric`, `CrsTransformerProj.isTargetMetric`
  - moved to #597

bekakh:
Update the code to use information from the updated CRS implementation (e.g. units per axis) described in https://github.com/interactive-instruments/xtraplatform-spatial/pull/87.

Closes #439.
Closes #398.


bekakh:
@azahnen Couldn't test the changes because ldproxy composite build is not starting correctly for me, caused by multiple OSGI/Apache Felix errors during startup. Looks like PROJ-JNI packages couldn't be resolved/are missing.

bekakh:
@azahnen I'm getting the same errors during startup, regardless whether with composite build or not.

bekakh:
@azahnen The errors are gone, but for some reason during startup nothing happens after having successfully loaded the codelists.

azahnen:
@rkoblenzer I am pretty sure that I fixed the multithreading issues. I ran a long load test overnight without errors. Can you run your tests again to verify this?

@cportele I implemented the run-in-docker approach, so you should be able to test this. You don't have to do anything special, it should just work.

azahnen:
@rkoblenzer To prevent issues, besides a `docker pull` you should delete the `data/proj` and `data/tmp` folders. I upgraded PROJ to 8.2.0 and that did not work without deleting the folders for me.

cportele:
@azahnen 

> I implemented the run-in-docker approach, so you should be able to test this. You don't have to do anything special, it should just work.

I did a quick test and it seems to work based on the log. Two observations:

* I have to change all the connectorInfo with 'localhost' to 'host.docker.internal'. 
* http://localhost:7080/ does not work (connection refused) and I could not find a hint in the log how the connect. I can see `Step 5/6 : EXPOSE 7080`, but that does not seem to have an effect. Any idea?

Thanks

azahnen:
@cportele I expected these issues would not occur since `--network=host` is set. But obviously that is only supported on linux: https://docs.docker.com/network/host/
I will make some adjustements and let you know.

azahnen:
@cportele 7080 and 5005 for debugging should now be published.

cportele:
@azahnen - It works, thanks. I have to manually attach the debugger at the beginning and stop the docker container at the end, but that is fine.

One question on the approach: Is there a reason why we are using the old "unit-ri" (v1) and not "indriya" (v2)?

cportele:
OK, I think I can answer "Is there a reason why we are using the old "unit-ri" (v1) and not "indriya" (v2)" myself: Because this is what PROJ JNI uses, correct?

azahnen:
No, not that I know of. Just change it if you like.

cportele:
OK, I will give it a try.

azahnen:
@cportele Actually PROJ-JNI has no direct dependency on it, but geoapi does: https://search.maven.org/artifact/org.opengis/geoapi-parent/3.0.1/pom

It seems that geoapi was not updated since 2017, is it discontinued or rather is there a successor?

azahnen:
Ok, good to know that this is not a dead end.

azahnen:
@rkoblenzer I pushed the caching changes, so as far as I am concerned you could start testing.

rkoblenzer:
@azahnen Thanks. I will check this today. 

Note: The TransformException "Point outside of projection domain" is still occurring when the tile seeding starts:

```
DEBUG [2021-12-06 17:05:54,478] lika                     - Stacktrace:
! org.opengis.referencing.operation.TransformException: Point outside of projection domain
! at org.kortforsyningen.proj.Transform.transform(Native Method)
! at org.kortforsyningen.proj.Operation.transform(Operation.java:782)
! at de.ii.xtraplatform.crs.infra.CrsTransformerProj.transform(CrsTransformerProj.java:110)
```


, and this error:

`ERROR [2021-12-06 17:05:54,480] lika                     - Cannot convert bounding box (6.919100, 51.001400, 7.242100, 51.196500) from EpsgCrs{code=4326, forceAxisOrder=LON_LAT} to EpsgCrs{code=25832, forceAxisOrder=NONE}. Reason: Failed to transform bounding box corner point coordinate(s): BoundingBox{xmin=6.9191, ymin=51.0014, xmax=7.2421, ymax=51.1965, epsgCrs=EpsgCrs{code=4326, forceAxisOrder=LON_LAT}}
`

(I adjusted the extent values already to check whether the extent can indeed be the cause - I don't think so.)

@bekakh or @cportele , can I fix this by adjusting the configuration, somehow - any ideas?

azahnen:
@rkoblenzer I thought we decided that the error should be logged, but seeding should continue. Before, the seeding was canceled when the error occurred.

rkoblenzer:
@azahnen You're right, thanks! (I thought, the seeding process had stopped, but it is not true - it is still active after the error.)

rkoblenzer:
@azahnen My impression after some testing with multiple threads: very good! So far, I always get a valid response regardless of the number of concurrent threads. So, I am pretty sure (as you were today) that the multi-threading problems are gone.

cportele:
@rkoblenzer - This looks strange, I have not yet seen this message yet (and 6.919100/51.001400 or  7.242100/51.196500 are clearly within the domain of epsg:25832). Which configuration produces these errors?

rkoblenzer:
@cportele It was lika, with a few additional CRS's and zoom levels for seeding. I can provide the configuration tomorrow morning.

cportele:
@rkoblenzer - OK, with lika I understand the "point outside projection domain" messages, because part of NRW is west of 6°. However, it is essential that we can convert also coordinates that are outside of the strict CRS extent without errors as it is frequently done, at least in Germany. I am not familiar with PROJ though, so I do not know how to enable such conversions without errors.

I have not seen the error.

rkoblenzer:
@cportele - OK, so you don't need my configuration, ok. Just in case: I have checked the configuration, there are also additional tile sets enabled (but I assume, this is not the cause). My deployment is on aaaxs14 (:7090) in a docker container. Let me know if you need more information.

I get also errors when I try to get the response in e.g. EPSG:31467 (native is EPSG:25832):

`.../lika/v1/collections/flurstueck/items?bbox-crs=http%3A%2F%2Fwww.opengis.net%2Fdef%2Fcrs%2FOGC%2F1.3%2FCRS84&crs=http%3A%2F%2Fwww.opengis.net%2Fdef%2Fcrs%2FEPSG%2F0%2F31467&limit=1&offset=0&skipGeometry=false` (ERROR 500, DEBUG TransformException "Point outside of projection domain").

```
ERROR [2021-12-07 07:46:46,122] lika                     - Server Error with ID d59d992cef1ca9fd: IO operation was stopped unexpectedly after 0 bytes because of java.l
DEBUG [2021-12-07 07:54:57,648] lika                     - Stacktrace:
! org.opengis.referencing.operation.TransformException: Point outside of projection domain
! at org.kortforsyningen.proj.Transform.transform(Native Method)
! at org.kortforsyningen.proj.Operation.transform(Operation.java:782)
! at de.ii.xtraplatform.crs.infra.CrsTransformerProj.transform(CrsTransformerProj.java:110)
```

This works: `.../lika/v1/collections/flurstueck/items?bbox-crs=http%3A%2F%2Fwww.opengis.net%2Fdef%2Fcrs%2FOGC%2F1.3%2FCRS84&crs=http%3A%2F%2Fwww.opengis.net%2Fdef%2Fcrs%2FEPSG%2F0%2F25832&limit=1&offset=0&skipGeometry=false" -H  "accept: application/geo+json"`.

cportele:
@azahnen @bekakh

I am wondering what role [CoordinateTupleWithPrecision](https://github.com/interactive-instruments/xtraplatform-spatial/blob/master/xtraplatform-crs/src/main/java/de/ii/xtraplatform/crs/domain/CoordinateTupleWithPrecision.java) as well as `CrsTransformerProj.isSourceMetric` / `CrsTransformerProj.isTargetMetric` still have? This also seems to prohibit coordinate accuracy at, e.g., sub-millimeter level. 

Isn't this replaced by the [configuration option in FEATURES_CORE](https://github.com/interactive-instruments/ldproxy/blob/e056dbf7b1229326b4ff7aec79e0ca2126a3dd60/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ldproxy/ogcapi/features/core/domain/FeaturesCoreConfiguration.java#L86)?

azahnen:
@cportele Yeah, I think these are leftovers from the original implementation and shouldn't be needed anymore. I will verify that and clean up.

cportele:
@azahnen 

While going through the list of open tasks I also found this:

> I will verify that and clean up.

Just to check: Is this something you are working on or should I create a PR that you can review?

azahnen:
@cportele It is on my list, I will take care of it later today.

PR #519, opened 2021-10-16
JSON-FG

### LINKED ISSUE #499: add support for JSON-FG
see https://github.com/opengeospatial/ogc-feat-geo-json

cportele:
This PR adds support for the current draft of the JSON-FG extensions. It also adds a schema resource for the GeoJSON feature collections, see https://github.com/opengeospatial/ogcapi-features/issues/612.

closes #499

Follow-up actions:

* extend schemas with the components for the JSON-FG extensions


cportele:
Fully agreed. However, before we update the transformation pipeline we should wait until there is more feedback, if JSON-FG will continue with the proposed GeoJSON-fallback with "geometry" always in WGS84. I will create a new issue and add a TODO in the code. That issue/TODO should be resolved before the module can eventually move into "stable".

Issue #498, opened 2021-08-29
Support for Maps (based on Vector Tiles and Styles)


cportele:
Render PNGs on the server based on the configured styles and Vector Tiles, delivery via Tiles API.

The rendering software (likely tileserver-gl) would run in its own Docker container, communicating via HTTP, with access to the vector tiles.

Notes:

* Tiling scheme is WebMercatorQuad, maps, e.g., in ETRS89/UTM32 cannot be supported directly.
* For tiles with text labels, the labels are partly cut off at the tile border, but the effect can be significantly reduced with symbol-avoid-edges.

cportele:
These changes build on PR #492 which adds general support for bitmap tiles.

Map tiles in WebMercatorQuad will be supported using a single style. There is currently no specification (or conflicting drafts) how to support multiple styles, in particular, if the API also supports OGC API Styles, but the styles are not identical to the map styles. Supporting multiple map styles will have to wait until these issues have been resolved.

* Since map tiles are a concept that is defined in OGC API styles, we will add the capability in the TILES building block.
* In TilesConfiguration a new option `mapProvider` will be added. Currently there is only one subtype "TILESERVER".
  * The only configuration options beside the `type` are
    * `urlTemplate` (string, mandatory, no default): a URL template for the access to the map tile from the tileserver-gl deployment. The URL template must include the parameters `tileMatrix`, `tileRow`, `tileCol` and `format`;
    * `tileEncodings` (string array, default is ["PNG"]) with "PNG", "JPG" and "WEBP" as recognized values.
* Tiles can be provided on the dataset (multi-collection, base path is `/`, the landing page) or on the collection level (base path `/collections/{collectionId}`).
* If TILES is enabled for the dataset or a single collection AND a map provider has been configured, the following resources will be made available relative to the base path, currently with WebMercatorQuad as the only supported `tileMatrixSetId`:
  * `/map/tiles`
  * `/map/tiles/{tileMatrixSetId}` 
  * `/map/tiles/{tileMatrixSetId}/{tileMatrix}/{tileRow}/{tileCol}`
* `/map/tiles`: Return the same information as for `/tiles`, reduced to WebMercatorQuad and links adjusted to the map URIs.
* `/map/tiles/{tileMatrixSetId}`: Same, but for `/tiles/{tileMatrixSetId}`
[...]

PR #522, opened 2021-10-25
Map Tiles Implementation

bekakh:
Add support for map tiles, as described in https://github.com/interactive-instruments/ldproxy/issues/498.


cportele:
See https://github.com/interactive-instruments/ldproxy/issues/498#issuecomment-984330781 how to activate support in a configuration that already supports TILES.

PR #532, opened 2021-10-28
Routes

azahnen:
I created a repo `routes` in gitea with the configuration. To send a test route query just start the configuration and call the items endpoint. I added a static route query there for testing purposes that also demonstrates the usage.


azahnen:
@cportele The redundant INSERT was generated because RouteQuery.end was also added to RouteQuery.wayPoints, see the fix I pushed.
I also found the reason why intermediate waypoints are not working (duplicate ids), I will push a fix for that tomorrow morning.

azahnen:
@cportele Intermediate waypoints should work now with the fix in xtraplatform-spatial and the adjusted configuration. I also implemented the warmup query and enabled it in the configuration.

cportele:
@azahnen - Thanks, one question on the configuration: The service now refers to sourcePath 'id', but there is no such path in the provider. Is an update in the provider missing?

azahnen:
@cportele I am not sure what you mean, I updated the provider: https://gitea.ci.interactive-instruments.de/ldproxy/routes/commit/c2e92a1cb94bf7f8857be1df382057c125b351f5

cportele:
@azahnen - Duh, I shouldn't have made comments in parallel during a web-meeting. Forget my comment...

cportele:
@azahnen - I updated this to use the master branch of xtraplatform-spatial. Something is now broken. AFAICT, the `data` property that joins the additional properties of a segment are now missing. Can you have a look?

azahnen:
@cportele Please try again. This was a bug from the transactional changes that affected all joined queries, not just routes.

cportele:
@azahnen - Yes, that fixed it, thanks.

azahnen:
@cportele Please check if the map client still works as expected and merge if there are no issues.

cportele:
@azahnen - The map client works as expected, however, there is an issue with the default style that is used. The point should only be shown at zoom level 12+. Otherwise you will see basically just see thousands of points for longer routes at smaller scales, which is not appropriate.

For this we need two separate presentation layers in the style, a single style as it currently is does not work. Can we support that, too?

```
    {
        'id': 'overview',
        'type': 'line',
        'source': 'route',
        'layout': {
            'line-join': 'round',
            'line-cap': 'round'
        },
        'paint': {
            'line-color': '#00f',
            'line-width': 5
        },
        'filter': ['==', '$type', 'LineString']
    },{
        'id': 'segment',
        'type': 'circle',
        'source': 'route',
        'minzoom': 12,
        'paint': {
            'circle-radius': 6,
            'circle-color': '#00f'
        },
        'filter': ['==', '$type', 'Point']
    }
```

azahnen:
@cportele Please try again. I also fixed the duplicate basemap issue.

cportele:
@azahnen - Excellent, thanks. For some reason, I no longer get a popup when I click on the point features (or maybe it is outside of the viewport), but I will check that detail later and merge the PR now.

cportele (approved):

Issue #538, opened 2021-11-17
500 response to a filter "string > 0"


cportele:
The request https://t17.ldproxy.net/fns/collections/notam/items?filter=lower_limit%3E0 results in a 500 response, but it should be a 400 response (since lower_limit is a string) or a 200 response (if the server is forgiving and would do an implicit string conversion).

cportele:
#603 adds a general type check before filters are used in queries. 

The only missing type information is for the properties, which is derived from the feature schema. Since the CQL2 module has no dependency to the features module, the type information is passed as a string. In error messages, the known types from the schema definitions are used, not the types used internally in the CQL2 implementation.

There are still some additional tests that could be done. For example, not all spatial operations can operate on all geometry types. This has been added in #599.

PR #548, opened 2021-11-30
Fix HTTP error code for an invalid CQL filter

bekakh:
Change HTTP response code from 500 to 400 for CQL filters that compare string properties with integer values, as described in https://github.com/interactive-instruments/ldproxy/issues/538.


cportele (commented):
Are all PSQL exceptions at this stage really always a client error?

bekakh:
@cportele Good point, I would assume not. I will narrow the scope to be more specific.

Issue #432, opened 2021-04-27
Limitations of the CQL implementation


cportele:
The current implementation of CQL has the following limitations:

* No support for array predicates, see #431.
* Only the following temporal predicates are supported: ANYINTERACTS, TEQUALS, AFTER, BEFORE, DURING.
* BOOLEAN literal: The following tokens are not accepted: "T" | "t" | "F" | "f" | "1" | "0".
* IN: currently restricted to a property on the left side and literals on the right side, NOCASE is not supported.
* No support for functions or arithmetic expressions.

bekakh:
### List of changes compared to the current implementation:

#### Temporal geometry literals (instants, CQL-Text)
Current:
```
1969-07-20
1969-07-20T20:17:40Z
```
CQL2:
```
DATE('1969-07-20')
TIMESTAMP('1969-07-20T20:17:40Z')
```

#### Temporal geometry literals (instants, CQL-JSON)
Current:
```
"1969-07-20"
"1969-07-20T20:17:40Z"
```
CQL2:
```
{ "date": "1969-07-20" }
{ "timestamp": "1969-07-20T20:17:40Z" }
```

#### Temporal geometry literals (intervals, CQL-Text)
Current:
```
1969-07-16/1969-07-24
1969-07-16T05:32:00Z/1969-07-24T16:50:35Z
2019-09-09/..
```
CQL2:
```
INTERVAL('1969-07-16', '1969-07-24')
INTERVAL('1969-07-16T05:32:00Z', '1969-07-24T16:50:35Z')
INTERVAL('2019-09-09', '..')
```

#### Temporal geometry literals (intervals, CQL-JSON)
Current:
```
[ "1969-07-16", "1969-07-24" ]
[ "1969-07-16T05:32:00Z", "1969-07-24T16:50:35Z" ]
[ "2019-09-09", ".." ]
```
CQL2:
```
{ "interval": [ "1969-07-16", "1969-07-24" ] }
{ "interval": [ "1969-07-16T05:32:00Z", "1969-07-24T16:50:35Z" ] }
{ "interval": [ "2019-09-09", ".." ] }
```

#### New CQL-JSON structure

CQL text: `city='Toronto'`

Current:
```
{ "eq": [ { "property": "city" }, "Toronto" ] }
```
CQL2:
```
{
  "op": "=",
  "args": [
    { "property": "city" },
    "Toronto" ]
}
```

#### Spatial predicates

[...]

PR #579, opened 2022-01-14
Update to CQL2

bekakh:
Update ldproxy to CQL2. See the spec changes listed in #432.

* [x] Update filter module 
* [x] Update integration tests 
* [x] Update bbox/datetime query parameters 
* [x] Disable JSON support (for now)
* [x] GPKG support

For follow-up tasks, see #599.


azahnen (approved):
I fixed the array and position tests. As far as I am concerned, this can be merged.

Issue #411, opened 2021-03-29
Transaction enhancements


cportele:
These are follow-up actions on #408.

API behaviour:

* [ ] Implement option to return an error, if PUT tries to create a new feature (this should be the default behaviour as it is the current behaviour specified in Part 4
* [ ] Implement PATCH

Schemas:

* [ ] Implement the create/replace/update schemas
* [ ] Add option to validate payload against the schema first
* [ ] Improve error messages. A 400 is now returned for most JSON or SQL errors including the exception message. We should "translate" at least the SQL errors.

Documentation:

* [ ] Update the documentation how to configure the security aspects (users, tokens)
* [ ] Document differences in the behaviour, if the featureId is a db-managed value or if it is taken from the data (a thematic identifier is used for the featureId)

PR #591, opened 2022-02-06
fix transactions

azahnen:


azahnen:
@cportele I tested this with zoomstack and daraa.

cportele:
@azahnen - I have tested it:

* It works for flat table features with no joins that use the primary integer key as the ID.
* For other IDs, PUT and DELETE result in SQL errors. Multiple POSTs create multiple features with the same ID.
* Features with joins result in SQL errors.

I will run a few more tests tomorrow. Maybe we should also add validation checks that disable the module, if the preconditions are not met for a collection?

azahnen:
> * It works for flat table features with no joins that use the primary integer key as the ID.
> * For other IDs, PUT and DELETE result in SQL errors. Multiple POSTs create multiple features with the same ID.

We only ever supported auto-incrementing primary keys, which are integers by nature.

> I will run a few more tests tomorrow. Maybe we should also add validation checks that disable the module, if the preconditions are not met for a collection?

Yes, that should be part of `typesValidation`, which would then set `provider.supportsTransactions` to disabled, which can then be handled in the transactions module. But that would be future work.

cportele:
I have added the follow-ups to #411.

cportele (approved):

PR #603, opened 2022-02-11
Add type checks for CQL2 expressions

### LINKED ISSUE #538: 500 response to a filter "string > 0"

cportele:
closes #538

see also https://github.com/interactive-instruments/xtraplatform-spatial/pull/109


azahnen (approved):
LGTM

PR #606, opened 2022-02-14
Update to JTS 1.18.2

cportele:
* Use the new `GeometryFixer` class in the TILES module and remove custom code that tries to fix invalid geometries.
* Remove custom code and options.


azahnen (approved):
LGTM

PR #615, opened 2022-03-02
add sortables, improve parameters and configurations

cportele:
This adds support for the Sortables endpoint in the jpms branch.

Testing also showed that the `isEnabledForApi()`-logic for some API components that are sub-resources of a collection was not working properly and were reported as disabled. For example, if sorting was enabled for specific collections, the parameter `f` was not considered enabled for the Sortables endpoint. The logic has been updated, the end-to-end tests have been updated to test with content negotiation and `f` against all endpoints.


azahnen (approved):
LGTM

PR #627, opened 2022-03-20
add FlatGeobuf support

cportele:
This pull request adds support for [FlatGeobuf](https://github.com/flatgeobuf/flatgeobuf). 

Additional changes as part of the edits:

* Feature and property classes for formats that only support Simple Feature geometries and "flat" attributes have been added (moved from the Tiles module). Any attributes with array or object values are converted to a JSON text representation of the array or object.
* `FanOutArrays` is a new schema visitor that updates a flattened schema to use property names identical to those in the instances.

While GeoJSON has much better support, the FlatGeobuf format has performance benefits. If I add a FlatGeobuf file (10.9 MB) with [all vineyards](https://demo.ldproxy.net/vineyards/collections/vineyards) in QGIS, the data is immediately there. Adding the equivalent GeoJSON file (15.3 MB) takes several seconds. (I have no idea why the difference in the user experience is so drastic; maybe because the FlatGeobuf file includes a schema and in the GeoJSON case the schema has to be guessed.)

The median time per request to access all vineyards 100 times (with 8 parallel requests):

* FlatGeobuf: 7.2s
* GeoJSON: 12.7s

Closes https://github.com/interactive-instruments/ldproxy-project-management/issues/46


azahnen (approved):
LGTM

PR #630, opened 2022-03-30
changing data

azahnen:
This is a reimplementation of #487:
- every entity now has `ChangingData`, which replaces `CollectionDynamicMetadataRegistry` (see interactive-instruments/xtraplatform#88)
- general pub/sub support for feature changes was added to providers (see interactive-instruments/xtraplatform-spatial#115)
- postgresql listen/notify support was added as an extension in a separate module since it is deemed experimental (see interactive-instruments/xtraplatform-spatial#115) (Hint: the channel has to be `{featureType}_{tableName}`, e.g. `notam_notam`)
  ```
  extensions:
    - type: FEATURE_CHANGES_PG
      enabled: true
      listenForTypes: # optional, default is to listen for all types
        - notam
  ```
- initialization of changing data and subscription to feature changes was moved to `CapabilityFeaturesCore`

**Todo**
* [x] Add a capability to periodically re-initialize the metadata from the feature data to account for lost notifications and/or because deletes cannot be not taken into account to update an extent. (See #633)
* [x] Publish feature changes on mutations in the provider. (Added to #411)
* [x] Discuss if the PostgreSQL listener should be removed when the two enhancements above are implemented. Are there any real-world cases where all three of these statements are true?
  - there are incremental changes to the database
  - these incremental changes are not applied through the API
  - up-to-dateness is so important that periodic recalculation is not sufficient
* [x] FeatureQueries.getFeatureCount() currently always returns 0, so for now the item count metadata element is not usable
* [x] The SQL feature provider also caches extent information, the whole data flow needs some consolidation.
* [x] Update documentation. (See #632)


cportele:
@azahnen - I tried the code with the T17 configuration/database (with a temporary hack to change the channel to just "notam", because this is the name in the database) and found the following issue: `SqlClientRx.getNotifications()` never returns any notification, because the Connection is a `PooledConnection`, not a `PGConnection`, and `PooledConnection` does not support PostgreSQL notifications.

azahnen:
@cportele Please try again.

cportele:
👍 Yes, that works.

PR #626, opened 2022-03-17
auto-docs

azahnen:
Closes #578.

PR #635, opened 2022-04-09
Update README.md

CabukEge:


azahnen:
Not needed, file will be generated when #626 is merged.

PR #642, opened 2022-04-12
return ETag header for a single feature response

cportele:
This will be required for conditional PUT/PATCH/DELETE requests.


azahnen (approved):
LGTM

PR #643, opened 2022-04-12
add itemCount to the Collection resources

### LINKED ISSUE #30: Add number of features on dataset pages?

cportele:
Where a count is set (and not negative) an "itemCount" member is added to the collection object in the extensions. It is also included in the HTML output.

closes #30 

@azahnen - I noticed that for more than half of the daraa collections, the value is set to -1, because of an exception when computing the count. For example, for CulturePnt. The exception is `org.postgresql.util.PSQLException: ERROR: variable not found in subplan target list`.

### DIFF
--- a/ogcapi-stable/ogcapi-collections/src/main/java/de/ii/ogcapi/collections/app/html/OgcApiCollectionView.java
+++ b/ogcapi-stable/ogcapi-collections/src/main/java/de/ii/ogcapi/collections/app/html/OgcApiCollectionView.java
@@ -33,6 +33,7 @@ public class OgcApiCollectionView extends OgcApiDatasetView {
 
     private final OgcApiCollection collection;
     public String itemType;
+    public final Optional<Long> itemCount;
     public boolean spatialSearch;
     public Map<String, String> temporalExtent;
     public List<String> crs;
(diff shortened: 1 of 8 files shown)


azahnen (approved):
I could not reproduce the issue, it works as expected for me.

Regarding the changes my only remark would be that the black number in the HTML views looks out of place. Maybe make it part of the link or use class `text-secondary`?

azahnen:
@cportele Googling the error message hints at a bug in current postgres versions: https://stackoverflow.com/questions/71126845/how-to-fix-internalerror-variable-not-found-in-subplan-target-list

I am using 9.6, so that would explain why it works for me.

cportele:
@azahnen - I was indeed using postgres 14 in that test, not 9.6. With the 9.6 docker images it also works on my machine.

I am improving the HTML of /collections and will then merge the PR.

azahnen:
@cportele Ok, great. 
From what I read the issue should occur for all current minor versions (14.2, 13.6, 12.10, 11.15, and 10.20). So we should test it again when the next bunch of versions is released.

cportele:
I will create an issue as a reminder.

PR #644, opened 2022-04-18
content disposition filename should be lower case

### LINKED ISSUE #638: http header Content-Disposition for queryables changed to upper case
The value of  http header `Content-Disposition` for `collections/id/queryables` was changed between release 3.1 and 3.2 from `inline; filename="id.queryables.html"` to `inline; filename="id.QUERYABLES.html"`. Like already discussed, it was not on purpose (should be in lower case).

cportele:
closes #638


azahnen (approved):
LGTM

PR #645, opened 2022-04-21
Typo

JakobMiksch:
Meanwhile there are three types of input data sources that are supported


cportele (approved):
Thank you, @JakobMiksch - good catch!

PR #650, opened 2022-04-28
fix link to ogc_api.yml

maaattes:
While replicating the demo, I noticed a broken link at [Step 4](https://github.com/interactive-instruments/ldproxy/tree/master/demo/vineyards#step-4-fine-tuning-the-api-configuration)


cportele (approved):
Thank you for submitting the fix!

PR #651, opened 2022-04-29
fix plugin deps

azahnen:
The gradle plugin added a dependency with version `+`, which introduced a bug in master without a code change.
Since the new dynamic version did not a match another static dependency, starting ldproxy failed with `Module com.google.common` not found.
This PR removes the dynamic version and fixes the issue.


cportele (approved):
Looks good, the issue is gone.

PR #662, opened 2022-05-21
Fix issues identified by the CITE tests

### LINKED ISSUE #660: OpenAPI document invalid
### ldproxy Version

3.2

### Current Behavior

OpenAPI: The schema of the `f` parameter for the Tilesets resource has an enum that includes "html" twice.

This issue was identified using the OGC CITE tests.

### Expected Behavior

Each enum value should be provided only once.

### Steps to Reproduce

Start any API with the Tiles module.

### Relevant log output

_No response_

### LINKED ISSUE #661: bbox requests with coordinates outside the native CRS domain return 400
### ldproxy Version

3.2

### Current Behavior

If the `bbox` parameter is provided with values outside of the domain of the native CRS, an error is thrown. An example is `bbox=-80,-5,-70,5` in the vineyards API which has ETRS89/UTM32N as the native CRS. 

This issue was identified using the OGC CITE tests.

### Expected Behavior

The correct response would be to return no features, not an error. The bbox should be reduced to the intersection with the spatial extent of the data.

### Steps to Reproduce

Send a request with `bbox=-80,-5,-70,5` to the vineyards API. The result is a 400 response with "Coordinates cannot be transformed: [-80.0, -5.0]"

### Relevant log output

_No response_

cportele:

Closes #660.
Closes #661.

### DIFF
--- a/ogcapi-draft/ogcapi-features-custom-extensions/src/main/java/de/ii/ogcapi/features/custom/extensions/infra/infra/EndpointPostOnItems.java
+++ b/ogcapi-draft/ogcapi-features-custom-extensions/src/main/java/de/ii/ogcapi/features/custom/extensions/infra/infra/EndpointPostOnItems.java
@@ -212,7 +212,7 @@ public class EndpointPostOnItems extends EndpointSubCollection {
                     throw new BadRequestException(result.get());
             }));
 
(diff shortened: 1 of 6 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-662.surge.sh

PR #672, opened 2022-06-10
Apple silicon

azahnen:
Support builds on Apple Silicon.

### DIFF
--- a/.woodpecker/release.yml
+++ b/.woodpecker/release.yml
@@ -10,7 +10,7 @@ pipeline:
   libs:
     image: openjdk:11-slim
     commands:
-      - ./gradlew publish dockerDistTar -Prelease=true -PdeployUser=$SNAPSHOT_USER -PdeployPassword=$SNAPSHOT_PASS
+      - ./gradlew publish dockerCopy -Prelease=true -PdeployUser=$SNAPSHOT_USER -PdeployPassword=$SNAPSHOT_PASS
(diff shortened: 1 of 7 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-672.surge.sh

cportele:
Thanks, this has helped, but I still run into some issues. It took a while to find all the places where the target Java version 11 was overridden with Java 17 in the IDE configuration. Now the dockerBuild process completes and the container starts, but then "> Task :dockerRun" is the last message and nothing is written to the log. I will see, if I can find out more over the next days.

azahnen:
@cportele The 11 vs 17 IDE issues look unrelated to silicon. As for the empty log on run, I saw that as well sometimes when starting from the terminal, but other times it worked as expected. Not sure what is going on there.

cportele:
@azahnen 

> The 11 vs 17 IDE issues look unrelated to silicon.

Yes, this is just that in the Apple Silicon Intellij version everything is set/changed to Java 17 and I had to revert that in a number of places to make the Docker build succeed.

> As for the empty log on run, I saw that as well sometimes when starting from the terminal, but other times it worked as expected. Not sure what is going on there.

Indeed, I just tried t again and now the server starts as expected. I will merge this now, thanks!

cportele (approved):

PR #696, opened 2022-07-10
add GML support for SQL feature providers

### LINKED ISSUE #681: GML support for SQL feature providers

The current building block "Features GML" supports only WFS feature providers. The goal would be to provide a generic GML feature encoding for SQL feature providers that is similar to the GeoJSON encoding.

Due to the GML design, this requires a few additional configuration settings. Since most of the options are GML-specific, they are added in the configuration of the GML building block, not the feature schema in the feature provider. Only GML-independent settings are added to the feature schema.

The additional requirements for mapping the feature schema to XML and for configuration options are discussed below.

### Describe the Use Case

Support also existing GML application schemas, if the structure of the feature schema in the feature provider is closely aligned with the GML application schema.

## Namespaces

Every XML element will have and XML attribute can have an XML namespace. To improve readability of the XML documents, it should be possible to specify a namespace prefix for a namespace.

Common namespaces and prefixes should be pre-defined, e.g. `xs` (XML Schema 1.0), `gml` (GML 3.2), `sf` (OGC API Features Core 1.0 SF).

The additional namespaces that are used for the data need to be configured in the GML configuration with their prefixes. For example:  

```yaml
  applicationNamespaces:
    ns1: http://www.example.com/foo/0.1
    ns2: http://www.example.com/bar/0.1
```  

There should also be a default namespace to use for XML elements, if no other namespace is specfied. The value will be the namespace prefix. This namespace will also be declared as the default namespace of the XML document.

```yaml
  defaultNamespace: ns1
```  

If any application namespaces should be included in an `xsi:schemaLocation` attribute in the root element, the document URIs have to be provided. 

```yaml
  schemaLocations:
    ns1: http://www.example.com/foo/0.1/foo.xsd
    ns2: http://www.example.com/bar/0.1/bar.xsd
```  

Pre-defined namespaces will only be added, if `schemaLocations` has been configured and the namespace is used for the feature collection root element (to meet XML Schema validation requirements).

## Property names

By default, the GML property element will receive the property name from the feature schema, that is, the element will be in the default namespace. A different name can be set using the ["rename" transformation](https://docs.ldproxy.net/providers/details/transformations.html#transformations), which also supports using another namespace.

Example:

```yaml
  transformations:
    myProperty:
      rename: 'ns2:myProperty'
```  

## Properties as XML elements or attributes

Properties may be represented as an XML child element or an XML attribute (if the property is one of STRING, FLOAT, INTEGER, BOOLEAN) of the parent GML object element. We need a configuration option to control this for a property.

```yaml
  xmlAttributes:
    - myAttProperty
    - someProperty.myOtherAttProperty
```  

### Properties with a multiplicity > 1

GML array properties are very rarely used, so it seems sufficient to map all items in the array to a GML property element.

### Properties with role ID

These properties must be a direct property of the feature type. They will be mapped to the 'gml:id' attribute.

If the values violate the rule for XML IDs, e.g., if they start with a digit, we need an optional configuration option to add a consistent prefix to map all values to valid XML IDs. 

Example:

```yaml
  gmlIdPrefix: '_'
```  

### Geometry properties

These will be mapped to the corresponding GML 3.2 geometry (gml:Point, gml:MultiPoint with gml:pos; gml:LineString, gml:MultiCurve, gml:Polygon, gml:MultiSurface with gml:posList). No 'gml:id' attribute will be added, but the 'srsName' will be set.

If a GML profile requires specific mappings (e.g., use of gml:Curve instead of gml:LineString, GML 3.3 simple geometry) this will need to be extended, but for now this should be sufficient.

### Link properties

OBJECT properties with object type 'Link' will be mapped to a gml:Reference value with 'href' and 'title' attributes, if set. 

### Measure properties

We need to identify properties that have to be mapped to a gml:MeasureType value (or one of its subtypes). We will support two different options:

- An OBJECT property with object type 'Measure' and the properties 'value' and 'uom', which both must be present.
- A FLOAT or INTEGER property with an additional information about a fixed unit. Since this is not GML-specific, the unit information should be part of the feature schema. Example:

```yaml
  measuredHeight:
    label: Gebäudehöhe [m]
    sourcePath: measuredheight
    type: FLOAT
    unit: m
```  

A third option would be pairs of properties like 'measuredHeight' and 'measuredHeight_uom' (i.e., identical name with a suffix), but this could also be added later, if necessary.

## GML object elements

Unlike in the GeoJSON encoding, all object/data type instances are represented through an GML object element.

In the feature schema, a name can be provided for each object type including the feature type itself. By default, this name will be used for the unqualified name of the GML object element.

If the GML object element is not in the default namespace, we need a configuration option to assign a namespace to an object type.

Example:

```yaml
  objectTypeNamespaces:
    FeatureTypeA: ns2
    FeatureTypeB: ns2
    DataTypeA: ns2
    DataTypeB: ns2
```

There may also be cases, in particular when inheritance is used in the underlying application schema, where multiple object types are represented in the same table with an attribute that specifies the name of the feature/object type. In this case we need a capability to identify the property and map the values to qualified names for the GML object element. In the following example, '_type' is the feature property with three different values mapped to the qualified element name.

```yaml
  variableObjectElementNames:
    FeatureTypeA:
      property: _type
      mapping:
        typeA1: 'ns1:TypeA1'
        typeA2: 'ns2:TypeA2'
        typeA3: 'ns2:TypeA3'
```

## Feature collection

Various feature collection elements are in use and sometimes additional ones are specified in GML application schemas. The default has to be a sf:FeatureCollection with a property sf:featureMember as [specified by OGC API Features](https://schemas.opengis.net/ogcapi/features/part1/1.0/xml/core-sf.xsd), but we need a configuration option to change the element names and state whether this feature collection element supports the WFS 2.0 StandardResponseParameters (timeStamp, numberMatched, numberReturned; the default is 'false').

Example:

```yaml
  featureCollectionElementName: 'plan:XPlanAuszug'
  featureMemberElementName: 'gml:featureMember'
  supportsStandardResponseParameters: false
```

A feature collection can include a gml:boundedBy property. However, when this element is written, the bounding box of the response in unknown. Therefore, no gml:boundedBy property is added to the response. 

### Conformance declaration

If applicable, the GML conformance declarations for the OGC API Feature GML-SF conformance classes should be set, if there is at least one collection with the GML building block enabled. Assessing conformance of the GML-SF standard would require analyzing the GML application schemas. It seems easier to make this a configuration option.

Example:

```yaml
  gmlsfLevel: 2
```

* No GML conformance will be declared, if `featureCollectionElementName` or `featureMemberElementName` are set in the configuration for at least one collection.
* Otherwise: GML-SF0 conformance will be declared, if all collections with GML enabled have a value `0` for the `gmlsfLevel` option.
* Otherwise: GML-SF2 conformance will be declared, if all collections with GML enabled have a value `0`, `1` or `2` for the `gmlsfLevel` option.
* Otherwise: no GML conformance will be declared.

### Additional Information

This could either be added in the existing GML building block or in a new one.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #681.

All options specified in the issue are implemented, except `includeBoundedBy`. I have second thoughts about this option, this information would not really be useful and could also be considered misleading.

### Open tasks

- [x] Add German documentation.
- [x] Add support for `unit` in `FeatureSchema`.
- [ ] There is a problem with the sequence of the properties. In general, the order of the properties is as specified in the feature provider, but this is no longer the case, if a join is involved. This requires an additional transformer in xtraplatform-spatial that restores the property sequence as specified in  the feature schema. Alternatively, this could also be addressed in  in the GML feature encoder.
- [x] Resolve PMD error messages.
- [x] There is an error in the CI pipeline when the documentation is processed.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-696.surge.sh

cportele:
Something went wrong with merging the master branch after the crud changes, so I close this PR and delete the branch and create new ones.

PR #705, opened 2022-08-02
support /search endpoints and parameterized stored queries 

### LINKED ISSUE #755: Add Search building block

Add support for:

- queries to multiple collections in a single request
- stored queries including parameterized stored queries

For background see https://github.com/opengeospatial/ogcapi-features/tree/master/proposals/search.

### Describe the Use Case

- Queries to multiple collections in a single request can reduce the number of connections and simplify processing the result
- Stored queries can simplify the execution of queries
- Stored queries can be used to limit the query capabilities provided by an API

See the OGC Testbed-18 Filtering Service and Rule Set Engineering Report (to be published).

### LINKED ISSUE #876: Improve linkage between endpoints and the formats that they support
It is currently not always simple to determine how formats (interface `FormatExtension`) and endpoints (interface `EndpointExtension`) are associated. This should be streamlined:

- Each endpoint should only support operations that return the same response format (methods: GET, POST) and that accept the same request format (methods: PUT, POST, PATCH). The abstract response and request format class can both be determined from the endpoint. The default request formats are the response formats that support transactions.
- The information about the path patterns supported by a format is no longer needed and can be removed. The path patterns have been prone to errors and make it difficult to reuse a format in a new endpoint in another building block.

azahnen:
### Pull request checklist

-   [x] Updated documentation

- Closes #755.
- Closes #876. This change was necessary to be able to use the feature formats in the new SEARCH building block.
- This builds on interactive-instruments/xtraplatform-spatial#139, interactive-instruments/xtraplatform-spatial#183 and #714.


azahnen:
@cportele You might use this branch to develop the search module or just merge it into your branch and close it.

azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-705.surge.sh

cportele:
This is now ready for review.

Sample configuration: `t18`. Deployment: [t18.ldproxy.net](https://t18.ldproxy.net/?tags=Filtering)

There are three open tasks (that could also follow-up tasks after merging):

- [ ] The repository for the stored queries must be moved to xtraplatform-spatial.
- [ ] To be able to document the stored queries in the OpenAPI definition, a capability is needed to update API definition one a stored query is updated at runtime. This is not an issue, if queries are provided as part of the configuration.
- [ ] There is a general issue with setting JSON Schemas for classes. In general, the right place should be `onStartup()`, but this method is only executed *after* the API definition has been compiled.

In addition there are known limitations with respect to parameterized stored queries (see documentation), but these can be addressed, if the need arises.

azahnen:
@cportele Since obviously I am the PR author and did all of the work, I can't approve it, so you will have to do that ;-)

This looks good to me, the format changes look reasonable. I made some tests with the predefined stored queries and some ad-hoc queries, it works as expected.

Regarding the open issues:

> The repository for the stored queries must be moved to xtraplatform-spatial.

I will tackle that along with #877, so stored queries will either go to the blob or entity store. 
Since they have an id and are used by the application and not just passed through, conceptually they should be auxiliary entities like codelists and tile matrix sets (#845). I will just have to check if/how that works with the user-defined stored queries. 
A side-effect would be that predefined stored queries could be written in YAML in addition to JSON, which might be more user-friendly.

 > To be able to document the stored queries in the OpenAPI definition, a capability is needed to update API definition one a stored query is updated at runtime. This is not an issue, if queries are provided as part of the configuration.

 > There is a general issue with setting JSON Schemas for classes. In general, the right place should be onStartup(), but this method is only executed after the API definition has been compiled.

I think both should be tackled along with #880 and #814, which will require a complete overhaul of the schema generation and caching anyways. You might add these requirements to the existing issues.

P.S.: I think we shortly discussed joins in multi collection queries before the testbed, which I guess would come in handy for the orchestration and eager loading use case. Was that discussed any further, might it be part of an additional requirements class or standard?

cportele:
@azahnen 

> Since obviously I am the PR author and did all of the work, I can't approve it, so you will have to do that ;-)

It's called a feature 😉 

OK, I will merge, thanks for the review.

> Regarding the open issues:

Sounds good. Being able to write the queries in YAML would indeed be nice.

> I think we shortly discussed joins in multi collection queries before the testbed, which I guess would come in handy for the orchestration and eager loading use case. Was that discussed any further, might it be part of an additional requirements class or standard?

Yes, we did. We did not get to that topic during the testbed, also because it was not necessary for the scenarios. But joins are mentioned as a future work topic in the Engineering Report and it is still on the agenda of the SWG. The idea is to list multiple collections in the query and have a capability to identify which property reference is associated with which collection.

cportele (approved):
See also https://github.com/interactive-instruments/ldproxy/pull/705#issuecomment-1427038388

PR #757, opened 2022-11-03
do not write "crs" member in GeoJSON

cportele:
### Pull request checklist

-   [ ] Updated documentation

A "crs" member is added to GeoJSON features when a different CRS is used. The content of the member differs from the original GeoJSON "crs" member and can confuse clients. It is safer to omit the member. 

JSON-FG provides the "coordRefSys" member to declare the CRS. Use JSON-FG and this member instead.

However, this is a breaking change (even though it is unclear, if any API user is using the "crs" member at the moment). So this change should only be introduced in a version 4.0.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-757.surge.sh

PR #761, opened 2022-11-03
bbox parameter: make property configurable

cportele:
### Pull request checklist

-   [x] Updated documentation

- [x] Add new configuration option `bboxProperty` in Features Core to identify a specific property to use for the "bbox" query parameter.
- [ ] Add new configuration option `datetimeProperties` in Features Core to identify a specific property to use for the "datetime" query parameter (instant or interval from two instants, or "null" for unbounded interval ends).

### DIFF
--- a/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/app/FeaturesQueryImpl.java
+++ b/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/app/FeaturesQueryImpl.java
@@ -347,9 +347,28 @@ public class FeaturesQueryImpl implements FeaturesQuery {
 
     Optional<FeatureSchema> featureSchema = providers.getFeatureSchema(apiData, collectionData);
 
-    featureSchema
-        .flatMap(SchemaBase::getPrimaryGeometry)
-        .ifPresent(geometry -> queryables.put(PARAMETER_BBOX, geometry.getFullPathAsString()));
+    apiData
+        .getExtension(FeaturesCoreConfiguration.class, collectionData.getId())
+        .map(FeaturesCoreConfiguration::getBboxProperty)
+        .ifPresentOrElse(
+            bboxProperty ->
+                featureSchema
+                    .flatMap(
+                        schema ->
+                            schema.getAllNestedProperties().stream()
+                                .filter(
+                                    p ->
+                                        p.getFullPathAsString().equals(bboxProperty)
+                                            && p.isSpatial())
+                                .findFirst())
+                    .ifPresent(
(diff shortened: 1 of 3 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-761.surge.sh

cportele:
@azahnen - After some more thoughts, I propose to close this PR (and #762) in favor of another approach.

The primary geometry property should in general be a proper Simple Feature geometry. Otherwise most feature formats (in particular GeoJSON, MVT, FlatGeobuf) cannot be supported with geometries that clients can deal with. In addition, I found that GEOS sometimes throws errors when executing `ST_Intersects` on a building polyhedron encoded as a `MultiPolygon`. For such geometries `ST_3DIntersects` needs to be used.

The only current use case for having special treatment of a second geometry property is JSON-FG, in case the feature has a polyhedron geometry that should be in the `place` member (and a footprint or representative point in the `geometry` member). 

My proposal would be:

* Continue to always use the primary geometry for `bbox`.
* Add a role `SECONDARY_GEOMETRY` in the feature schema. If set, this spatial property will be used for the `place` member, otherwise the primary geometry.
* For cases where the primary geometry is a polyhedron (i.e., a multi-polygon that has the constraints `composite` and `closed` in the schema), use `ST_3DIntersects` instead of `ST_Intersects` for PostGIS. The information to use a 3D spatial operator, where available, could be added by ldproxy to the `FeatureQuery`.
* During startup, add checks for all feature formats that only support Simple Feature geometries that the primary geometry is a "proper" Simple Feature geometry.
* Add the code from #762 to make invalid geometries in the FGB/MVT encoders valid.
* Potentially add a new value transformer for spatial properties `reduceToPoint` that maps a geometry to a point (could be the centroid or just any point).

Any thoughts?

azahnen:
@cportele In general this seems to be a much cleaner approach. There might be other encodings where more than one geometry plays a special role, so introducing a new feature schema role feels like the natural solution.
One comment:
> The information to use a 3D spatial operator, where available, could be added by ldproxy to the FeatureQuery.

I think I would prefer more flags or additional types in the feature schema if needed to identify e.g. a polyhedron. The spatial operators are an implementation detail of PostGIS, other backends might not need this 2d/3d distinction. So I think this should be handled completely in the provider and not controlled by an upper layer.

cportele:
Good, I implement this approach, except for ST_3DIntersects part.

> So I think this should be handled completely in the provider and not controlled by an upper layer.

That makes sense. The information that we have a polyhedral surface is already encoded in the schema, but we need a mechanism to pass this information to the SQL filter encoder.

azahnen:
> but we need a mechanism to pass this information to the SQL filter encoder.

I can take a look at that.

azahnen:
@cportele See interactive-instruments/xtraplatform-spatial#190. Let me know if that makes sense.

PR #762, opened 2022-11-03
Geometry improvements to the feature encodings with flat attributes (mvt, fgb)

cportele:
### Pull request checklist

-   [x] Updated documentation

Update the MVT and FlatGeobuf formats:

- support a configuration option `geometryProperty` to use a geometry property that is not the primary one
- try to make invalid geometries valid

Depends on changes in the `bbox-property` branch.

Potentially split into two pull requests and separate the changes that try to make invalid geometries valid. This does not depend on the bbox-property changes.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-762.surge.sh

cportele:
Replaced by #916

PR #766, opened 2022-11-03
skip f parameter in self links, if there is no alternate link

cportele:
### Pull request checklist

-   [ ] Updated documentation

In "self" links, the "f" parameter should not be set, if there is no alternate encoding (and maybe also no parameter "f").


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-766.surge.sh

PR #769, opened 2022-11-03
add support for CSV output

cportele:
### Pull request checklist

-   [ ] Updated documentation

In Testbed 18 there was a use case for a CSV output. Is this something that could also be valuable in other contexts, too?


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-769.surge.sh

azahnen (changes requested):
I made a short test and it seems that properties are not flattened by default, instead the CSV columns contain JSON.
Bildschirm­foto 2022-11-11 um 09 42 42

I guess CSV should always be flattened like tiles, so `CsvConfiguration` would need [this](https://github.com/interactive-instruments/ldproxy/blob/2d11194ce49dcad39f3cd458e9f5c0ecafaeb824/ogcapi-draft/ogcapi-tiles/src/main/java/de/ii/ogcapi/tiles/domain/TilesConfiguration.java#L715).

cportele:
@azahnen - Can you have another look, I introduced the changes that we discussed.

There is one [TODO in the code](https://github.com/interactive-instruments/ldproxy/blob/features-csv/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/domain/FeatureEncoderSfFlat.java#L34-L38). Two of the three encoders included this code, one did not. I tried it without the code and everything seemed to work as expected. Can we just delete the code or when is it necessary?

azahnen:
@cportele Regarding the TODO, the code is only needed when the encoder actually writes to the output stream. That does not seem to be the case e.g. for `FeatureEncoderFlatgeobuf`, it is using `push` directly. So I think it can be safely removed.

The `OutputStreamToByteConsumer` is used as a bridge for third party libs like Jackson or Mustache that expect an actual `OutputStream` to write to.
I think in the end the `OutputStream` should be removed completely from transformation contexts, it is a leftover from the previous iteration of the transformation engine. If the bridge is needed, it should be an implementation detail of the encoder.

cportele:
Ah ok, I understand. I will remove the code then. All three encoders use push(), the outputStream is not used.

azahnen (approved):
LGTM

PR #778, opened 2022-11-15
Issue 775

### LINKED ISSUE #775: configuring TILES in overrides
### ldproxy Version

master

### Current Behavior

If building block TILES is configured in store/entities/services, the tiles configuration for the api and each collection is as expected.

If the building block TILES configuration is moved to store/overrides/services, the tiles configuration for the api is the same, but the configuration for each collection is the default configuration (enabled=false, etc.).

There is also the following behavior, which may be related?

If I use the following configuration in store/entities/services, it works. If I use it with the ldproxy-cfg.jar, there is an error: "Missing type id when trying to resolve subtype of [simple type, class de.ii.ogcapi.tiles.domain.TileProvider]: missing type id property 'type' (for POJO property 'tileProvider')"

```yaml
- buildingBlock: TILES
  enabled: true
  transformations:
    '*':
    - flatten: "."
  tileProvider:
    type: FEATURES
    center:
    - 8.589
    - 50.117
  cache: MBTILES
```

### Expected Behavior

There is no difference whether the building block TILES is configured in entities or overrides.

### Steps to Reproduce

Move the TILES building block in the daraa configuration to overrides. In the multi-layer seeding all collections are then skipped, because no collection is enabled.

### Relevant log output

_No response_

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #775.

### DIFF
--- a/ogcapi-draft/ogcapi-tiles/src/main/java/de/ii/ogcapi/tiles/app/TileProviderFeatures.java
+++ b/ogcapi-draft/ogcapi-tiles/src/main/java/de/ii/ogcapi/tiles/app/TileProviderFeatures.java
@@ -63,7 +63,8 @@ public abstract class TileProviderFeatures extends TileProvider {
    * @langDe Fester Wert, identifiziert die Tile-Provider-Art.
    * @default `FEATURES`
    */
-  public final String getType() {
+  @Value.Derived
+  public String getType() {
     return "FEATURES";
   }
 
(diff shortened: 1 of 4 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-778.surge.sh

cportele (approved):
I have verified that the changes fix both issues, thanks

PR #783, opened 2022-11-17
use buffer around the data bbox in tile queries

### LINKED ISSUE #709: filters with degenerate envelopes return no features
### ldproxy Version

master

### Current Behavior

If a query includes a bbox where the corner points are the same or where at least one coordinate is the same, the result is empty. Such geometries may be considered invalid polygons in backends, e.g. PostGIS. 

Note that this also affects, e.g., a collection with a single feature with point geometry. Since the bbox is reduced to the extent of the collection, any bbox query, including for tiles, is reduced to a degenerate bbox that is a point.

### Expected Behavior

In OGC API Features the geometry operation is expected to be executed with the proper geometry.  That is, an envelope that is a point has to be mapped to a point and an envelope that is a horizontal/vertical line has to be mapped to a line string.

### Steps to Reproduce

Create a collection with a single point feature and generate the vector tiles for the collection. They are all empty including where the feature is.

### Relevant log output

_No response_

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #709.

In FilterQuery we already use a buffer around the extent of the data to avoid unexpected behavior because of floating point issues in case of collections with a single point feature or when converting a bounding box to the native CRS. We also need this for tile queries.

Note that this is a PR again the `redesign-tiles` branch.

### DIFF
--- a/ogcapi-draft/ogcapi-tiles/src/main/java/de/ii/ogcapi/tiles/app/TileFormatMVT.java
+++ b/ogcapi-draft/ogcapi-tiles/src/main/java/de/ii/ogcapi/tiles/app/TileFormatMVT.java
@@ -76,6 +76,8 @@ import org.slf4j.LoggerFactory;
 public class TileFormatMVT extends TileFormatWithQuerySupportExtension {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(TileFormatMVT.class);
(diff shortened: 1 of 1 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-783.surge.sh

PR #784, opened 2022-11-20
Add option to omit large Link header

cportele:
### Pull request checklist

-   [x] Updated documentation

Large HTTP headers increase the size of the response and may cause an exception in dropwizard. This adds a `linkHeaderLengthLimit` option to omit the Link header, if the text length of the link is larger than the limit.

No default limit value is set for backwards compatibility. In version 4.0 we should consider to set a default limit.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-784.surge.sh

azahnen:
@cportele After further consideration, I think a cfg option does more harm than good. 
I removed the option, instead the limit is always one fourth of `maxResponseHeaderSize`. So with the default settings the limit is still 2048 bytes.
I added the computed value to `ApiRequestContext` for convenience so that it does not have to be passed into `prepareSuccessResponse` on every call.

azahnen (approved):

PR #759, opened 2022-11-03
improve content negotiation

cportele:
### Pull request checklist

-   [ ] Added/updated unit tests
-   [ ] Updated documentation
-   [x] All checks are passing

### Changes introduced by this PR

This PR makes content negotiation for the media type more robust. In particular, content negotiation now supports media type parameters. This PR is required, for example, for the GeoJSON compatibility mode of JSON-FG added by #758.

PR #796, opened 2022-11-26
fix ApiMediaType::matches

cportele:
### Pull request checklist

-   [ ] Updated documentation

ApiMediaType::matches used content negotiation, but must just check compatibility, including parameters (which MediaType::isCompatible ignores). The target media type has already been negotiated.

This is a follow-up to #759. Before that pull request, the behaviour was stricter.

The pull request also improves a related error message.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-796.surge.sh

PR #797, opened 2022-11-26
entity groups

### LINKED ISSUE #773: introduce entity groups

Introduce entity groups which share common defaults.

### Describe the Use Case

Currently deployment-wide defaults for entities can be set in `store/defaults`. These are applied to all entities in `store/entities`.
It would be helpful to have defaults that are only applied to a subset of entities.

This feature introduces hierarchical entity groups which share common defaults. Groups are defined by using additional directories in the store. 
So for example defaults defined in `store/defaults/group1` will be applied to entities in `store/entities/group1`.

Multiple subdirectories can be used for sub-grouping and all defaults will be merged in order to avoid redundancy.
For example a given provider defined in `store/entities/group1/subgroup1/providers/provider1.yml` will use the following defaults:
- application defaults
- deployment defaults in `store/defaults/provider.yml`
- group1 defaults `store/defaults/group1/provider.yml`
- subgroup1 defaults `store/defaults/group1/subgroup1/provider.yml`

### Additional Information

Entity ids would still have to be unique per entity type. Otherwise we would have to add the groups as prefix to the id, which would change the URL for services as an unwanted side-effect.

Overrides have to match the entity groups, so for the example above overrides would have to reside in `store/overrides/group1/subgroup1/providers/provider1.yml`.

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #773.

Also see interactive-instruments/xtraplatform#108


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-797.surge.sh

cportele (approved):
LGTM. Tested with demo and groups "tiles" (openmaptiles), "features" ©hapes, geoinfodok) and "features/tiles" (daraa, strassen, vineyards, zoomstack).

PR #801, opened 2022-11-28
Tile improvements

### LINKED ISSUE #779: Issues with MBTiles tile providers
### Discussed in https://github.com/interactive-instruments/ldproxy/discussions/746

<div type='discussions-op-text'>

<sup>Originally posted by **ghelle-fhh** October 26, 2022</sup>
Hello,

I was attempting to set up an API where only an MBTiles file is the provider and OGC API Features is not enabled, similar to the example on the demo site:
https://demo.ldproxy.net/openmaptiles

However I cannot seem to find any futher documentation or examples. It appears to be referenced in the documentation, but links lead to nowhere:

https://docs.ldproxy.net/services/building-blocks/tiles.html#tile-provider</div>

There are two issues:

* The spatial extent in an MBTiles tile provider is not used to determine the spatial extent of the tile set.
* If the spatial extent of the API cannot be determined, no tile matrix set is included in the HTML dropdown list.

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #779 

Tested with demo configurations plus in addition also a zoomstack API that provides no features, just tiles using the MBTiles file.

#### HTML

In cases where the API or collection has no associated spatial extent (e.g., an API with a MBTiles provider), the information for each tile matrix set was not passed to the HTML representation.

The initial view of a map is determined as follows: For feature views, always use the bounding box of the features. For tile views, the following sequence is used:

1. Use the style, if available, which will use the center/zoom specified in the style
2. Use center/zoom, if available (from the tileset, either from the configuration or from the MBTiles metadata)
3. Use bounding box, if available (from the tileset, either from the feature data or from the MBTiles metadata)
4. Use default spatial extent (only for tile views, from the API configuration)

#### TileSet resources

For MBTiles providers:

- include vector layers
- use bounding box from metadata

This still needs to be updated for the new tile providers.

### DIFF
--- a/ogcapi-stable/ogcapi-html/src/main/java/de/ii/ogcapi/html/domain/MapClient.java
+++ b/ogcapi-stable/ogcapi-html/src/main/java/de/ii/ogcapi/html/domain/MapClient.java
@@ -49,7 +49,16 @@ public interface MapClient {
     return false;
   }
 
+  List<Double> getCenter();
+
+  Optional<Number> getZoom();
+
   Optional<Map<String, String>> getBounds();
+
+  @Value.Default
+  default boolean getUseBounds() {
+    return false;
+  }
 
   @Value.Default
   default boolean drawBounds() {
--- a/ogcapi-draft/ogcapi-maps/src/main/java/de/ii/ogcapi/maps/app/MapTileSetsFormatHtml.java
+++ b/ogcapi-draft/ogcapi-maps/src/main/java/de/ii/ogcapi/maps/app/MapTileSetsFormatHtml.java
@@ -137,9 +137,6 @@ public class MapTileSetsFormatHtml implements TileSetsFormatExtension {
     return new TileSetsView(
         api.getData(),
         tiles,
-        collectionId,
-        api.getSpatialExtent(collectionId),
-        api.getTemporalExtent(collectionId),
         tileMatrixSets,
         breadCrumbs,
         requestContext.getStaticUrlPrefix(),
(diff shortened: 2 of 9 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-801.surge.sh

ogcapi-stable/ogcapi-html/src/main/java/de/ii/ogcapi/html/domain/MapClient.java
```diff
@@ -49,8 +49,17 @@ default boolean getSavePosition() {
...
   }
 
+  List<Double> getCenter();
+
+  Optional<Number> getZoom();
+
   Optional<Map<String, String>> getBounds();
 
+  @Value.Default
+  default boolean getUseBounds() {
```
azahnen, line 59: I think this is redundant and can be removed. I do not see a case in the code where it would make any difference. Even if that was the case, getBounds is already optional, so just do not set it if you do not want bounds.

azahnen (changes requested):

ogcapi-stable/ogcapi-html/src/main/java/de/ii/ogcapi/html/domain/MapClient.java
cportele: It is necessary. If bounds are included in _map, the JavaScript always moves the view to the bounds. This is desired for Features (useBounds=true), but not for Tiles (useBounds=false), so the bounds are suppressed for tiles, if we have a style or explicit center information. See mapClient.mustache.

azahnen (approved):

Issue #719, opened 2022-09-09
Transformations on geometry properties are ignored


cportele:
### ldproxy Version

master

### Current Behavior

Property transformations like `rename` or `remove` on geometry properties have no effect.

### Expected Behavior

Property transformations are processed also on geometry properties.

### Steps to Reproduce

Try to `rename` or `remove` a geometry property.

cportele:
Fixed by https://github.com/interactive-instruments/xtraplatform-spatial/pull/144

PR #805, opened 2022-12-02
Features HTML: improve hideMap behavior

cportele:
### Pull request checklist

-   [ ] Updated documentation

Previously transformations on geometry properties had no effect. This was fixed in #719. In the Features HTML representation, no map is shown when no geometry is present in the data. Since the bugfix, no map is now shown, if the Features HTML configuration included a transformation that removes the geometry property from the response. Such a transformation does not really make sense, but is present in some existing API configurations. To support those configurations, the behavior is changed to hide the map, if the feature schema (without transformations) includes a primary geometry property.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-805.surge.sh

azahnen (approved):
LGTM

PR #829, opened 2022-12-13
Use zoomLevelsCache

cportele:
### Pull request checklist

-   [ ] Updated documentation

zoomLevelsCache can be used to reduce the levels for which tiles are cached. If the option is not set for a tiling scheme, then tiles are cached at all levels. This changes uses getZoomLevelsDerived as the default, but in the presence of getZoomLevelsCacheDerived that range is used. Seeded levels are excluded. The assumption is that only higher levels are not seeded, if seeding is used.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-829.surge.sh

PR #865, opened 2023-01-05
fix schemaLocation edge cases

### LINKED ISSUE #864: GML: use of custom GML schemas breaks schemaLocation
### ldproxy Version

3.3.1

### Current Behavior

If the GML application schema imports the GML schema from a different location than the official URL in the OGC schema repository, this can result in validation errors depending on the parser, since two different GML schema documents are referenced.

### Expected Behavior

This does not lead to validation errors. To fix, this requires two changes:

* Reduce the namespaces in the schemaLocation attribute to the necessary namespaces. That is: the application namespaces plus the namespace of the root element.
* Support that the pre-defined schema locations, in particular the GML schema location, can be changed in the schemaLocations configuration option (currently the result is an error due to a key conflict). If configured, the canonical schema location would be overwritten.

### Steps to Reproduce

Reference a different GML schema or a GML profile from the GML application schema and validate any GML document created by the API. Note that validators may handle this situation without an error, but depending on their strategy they may throw an error (e.g. XmlSpy).

### Relevant log output

_No response_

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #864.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-865.surge.sh

PR #641, opened 2022-04-11
Updates while editing/reviewing the release candidate of Part 3

cportele:
This partly addresses items in #599. Current changes:

* Add support for additonalProperties in the Queryables resource; this requires support for Boolean JSON Schemas. Note: Originally there was an issue with the serialization of `additionalProperties`. It returned `"additionalProperties": [ "ImmutableJsonSchemaFalse", false ]` instead of `"additionalProperties": false`. This was fixed by the changes described in the next comment.
* Queryables: Fix issues in the HTML representation of the data type of spatial and temporal queryables.
* See also https://github.com/interactive-instruments/xtraplatform-spatial/pull/118

PR #867, opened 2023-01-12
fix json schema serialization + deserialization

### LINKED ISSUE #659: Fetching style metadata can lead to server errors
### ldproxy Version

master

### Current Behavior

The metadata of a style cannot be accessed, an exception is thrown. At least for cases where the metadata is not derived on-the-fly, but where a metadata resource is on the server. This is a result of a [recent change](https://github.com/interactive-instruments/ldproxy/pull/641#issuecomment-1094943095). The style metadata includes JSON Schemas, so we need to be able to deserialize it. After the removal of the `@JsonTypeInfo` annotations, Jackson no longer knows which subtype to create. 

Note that most deployments will only derive the metadata on the fly, so this will affect only a few deployments (like the Daraa demo API, but we can remove the specific metadata for now).

### Expected Behavior

The style metadata is returned with a 200 response.

### Steps to Reproduce

Access any stored resource `/styles/{id}/metadata` where the resource in the server repository.

### Relevant log output

```shell
Server Error with ID 534fb00ca9fdf0ef: Cannot construct instance of `de.ii.ogcapi.features.core.domain.JsonSchema` (no Creators, like default construct, exist): abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: de.ii.ogcapi.styles.domain.ImmutableStyleMetadata$Json["layers"]->java.util.ArrayList[7]->de.ii.ogcapi.styles.domain.ImmutableStyleLayer$Json["attributes"]->java.util.LinkedHashMap["F_CODE"])
```

azahnen:
### Pull request checklist

Closes #659.

- reverts the changes from #641 
- addresses the original issue by using custom converters for `JsonSchemaTrue` and `JsonSchemaFalse`
- enables deserialization for most subtypes, I am pretty sure it does not work for subtypes without an actual `type`


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-867.surge.sh

azahnen:
Postponed to v3.4, rebased to master.

azahnen:
Deserialization should work for all subtypes before merging.

cportele:
@azahnen 

I have changed the JSON Schema deserialization to a custom converter to also handle the schemas without a `type` property (e.g., `oneOf`, `$ref`, `true`, `false`).

In addition, this fixes two issues in the Styles module that I noticed during testing.

Could you review, if this is ready for merging?

PR #912, opened 2023-02-22
Filter editor

### LINKED ISSUE #899: improve HTML filter editor
The main focus is to improve the user experience, but we should also look at maintainability. This is the currently planned list of actions:

- [x] make it easier to adjust already applied filters
- [x] fetch inputs from API endpoints instead of injecting them into the HTML
- [x] use selects for fields with enums/codelists
- [x] hide irrelevant parts of UI, i.e. bbox filtering when collection has no geometry
- [x] use number fields when appropriate
- [x] ~evaluate better options for the datetime picker~ (we decided to stay with the current solution)
- [x] custom hooks refactoring
- [x] submit on enter
- [x] context sensitive buttons
- [x] date range slider in addition to datetime pickers (see https://developers.arcgis.com/javascript/latest/sample-code/timeslider-filter/)
- [x] enter datetime via keyboard

PZahnen:
### Pull request checklist

-   [ ] Updated documentation

Addresses #899 .


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-912.surge.sh

cportele (approved):
Everything works as expected. I only noticed one minor styling "issue": The blue blocks of selected filters are spaced differently whether the filter editor is active or not (if I have two filters, there is a small gap, if inactive, and no gap, if active).

PR #917, opened 2023-02-27
Codelist schema

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Currently codelists are only used in the returnables schema, but actually in a wrong way. The enum uses the code list keys instead of the code list values.
This fixes the returnables schema and also adds codelists to the queryables schema, where the codelist keys are used for the enum.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-917.surge.sh

ogcapi-stable/ogcapi-features-geojson/src/main/java/de/ii/ogcapi/features/geojson/domain/SchemaDeriverReturnables.java
```diff
@@ -37,7 +37,7 @@ public SchemaDeriverReturnables(
       String label,
       Optional<String> description,
       List<Codelist> codelists) {
-    super(version, schemaUri, label, description, codelists);
+    super(version, schemaUri, label, description, codelists, false);
```
cportele, line 40: The value in the returnables is the key (unless there is a codelist transformation already in the provider, not "just" in Feature HTML). So, `false` needs to be `true` like for the queryables and sortables.

ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/domain/SchemaDeriverJsonSchema.java
```diff
@@ -314,7 +317,10 @@ protected JsonSchema withConstraints(
             property.isArray()
                 ? property.getValueType().orElse(SchemaBase.Type.UNKNOWN) != SchemaBase.Type.INTEGER
                 : property.getType() != SchemaBase.Type.INTEGER;
-        Set<String> values = codelist.get().getData().getEntries().keySet();
+        Collection<String> values =
+            useCodelistKeys
+                ? codelist.get().getData().getEntries().keySet()
+                : codelist.get().getData().getEntries().values();
```
cportele, line 323: The use of `values()` can lead to exceptions. This is a follow-on effect from my other comment. If there is an integer key, `string` will be false, but the `values` are proper strings and the `Integer.valueOf(val)` below will throw a NumberFormat exception.

Example case: https://demo.ldproxy.net/daraa/collections/TransportationGroundCrv/schemas/feature

cportele (changes requested):
Adding the codeelist values to the queryables is correct, but this statement is not correct:

> Currently codelists are only used in the returnables schema, but actually in a wrong way. The enum uses the code list keys instead of the code list values.

In general, the key is the value in the data and also in the returnables. A capability would be helpful to express both the key and the value in the JSON Schema, but this is not possible with the `enum` keyword. 

I typically include a link to the codelist in the description, if it is pulished somewhere, but of course this is not helpful for software processing the schema and also an extra step for human users.

As an aside, glTF actually supports to provide a description for the codes/keys in the data. Example: https://d123.ldproxy.net/nunsisland/collections/buildings/gltf/schema

For the filter editor: If Features HTML includes a `codelist` transformation for a property, then the most helpful way would be to list the value in the drop-down list and convert it to the key in the filter.

Related: Should we publish the codelists as separate resources under `/codelists/{codelistId}` in the API in JSON and HTML?

ogcapi-stable/ogcapi-features-geojson/src/main/java/de/ii/ogcapi/features/geojson/domain/SchemaDeriverReturnables.java
azahnen: The schema generation is based on constraints, it does not take into account value transformations. I would have assumed that a codelist constraint is accompanied by a codelist transformation, but that is obviously not the case in your example. 
We should further discuss this next week as part of the broader schema discussion I wanted to have.
I reverted the change for now.

ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/domain/SchemaDeriverJsonSchema.java
azahnen: Well, the use of `keySet()` can also lead to exceptions, so this was unsafe code before the PR. I pushed a change that filters out non-integers.

azahnen:
@cportele 
Regarding the filter editor, I agree with your proposal, but we do not have that information. That would require publishing codelists like you suggested, but also referencing them in the schema, e.g. with a custom `codelist` property along the lines of `unit`. I think we should discuss that in a new issue.

ogcapi-stable/ogcapi-features-geojson/src/main/java/de/ii/ogcapi/features/geojson/domain/SchemaDeriverReturnables.java
cportele: The codelist constraint in the feature schema only indicates that the value range should be from the list of codes in the codelist. Whether or where a codelist transformation is applied is another matter, typically only in the HTML representation. 
Agreed to discuss this in the broader schema context. We can mark this as resolved for this PR.

cportele (approved):

PR #929, opened 2023-03-09
Slider

### LINKED ISSUE #899: improve HTML filter editor

PZahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #899 .


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-929.surge.sh

cportele:
@PZahnen - I just tested the slider, looks good. 

One comment: for "instant" the slider looks like an interval from -infinity to the dot (left is blue, right is grey). For "instance" only the dot should be blue and the bar grey on both sides.

PZahnen:
@cportele - Thanks for the feedback. I have made the necessary adjustments to the slider.

cportele:
@PZahnen 

I tried it again and detected several issues when accessing the `items` endpoint, all using APIs from the `demo` configuration.

I tried the `cshapes` configuration. In the browser console, I see the error "Error: Slider range is too large.  Increase step value."

There is also some debug logging in the console that eventually needs to be removed.

With `daraa` and collection `CulturePnt` I get the following error in the browser console:

> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

With `vineyards` I can access the filter editor and the instant slider now works as expected, but the period slider does not work (no period and the dot is fixed on the left - this is probably the result of the fact that the temporal extent is an instant only).

PZahnen:
@cportele - The identified issues have been resolved successfully.

cportele:
@PZahnen  - Thanks!

Additional observation and improvement ideas after more testing:

## `cshapes`

The data type of the primary temporal properties in the feature schema is `DATE`. The Filter Editor, however, uses timestamps as if the data type would be `DATETIME` (both for the slider and the literal values). It would be better to use the data type as used in the schema.

When I select an "instant" using the slider, after selecting the instant using the blue button, the selection switches to "period". This should not happen.

In the BBOX map, I am not able to change the longitude values from -180 and +180 respectively. I can only change this in the numeric fields.

## `daraa`

For collections with one feature only (e.g. `PhysiographyCrv`), the filter editor includes the FIELD and BBOX filters, but not the DATE/TIME filter. 

The labels on the slider can be confusing. For collection `TransportationGroundCrv`, the tick labels are "Dec 2011", "Dec 2012", etc. For collection `RecreationPnt`, the labels are "Dec 11", "Dec 13", etc. Could the latter be written as "11 Dec", "13 Dec", etc. to be less confusing?

The "Period" slider for collection `RecreationPnt` covers the period from 9 Dec to 26 Dec correctly. When I change to instant, the slider changes to the period from 2011 to 2023. I see a similar behavior also in other collections, could it be that the end is always set to "now"?

## `strassen`

Collection `unfaelle`: In the literal representation, the timestamps are shown without a timezone, which might be misunderstood as the local time. In the filter `Z` is added as the timezone. Maybe either add the `Z` already  to the literal representation or simply change "DATE/TIME" to "DATE/TIME (UTC)"? (Another option could also be to convert timestamps to local time in the filter editor UI since this may be more intuitive for a user, but I would not make such a change now.)

cportele:
@azahnen 

> This information is currently not available in the filter editor and I also don't think there is a way to retrieve it with an api call, or is there?

No, there isn't. In a way this information is typically part of the queryables, but a) this does not have to be the case and b) the queryables endpoint is optional. What we could do - as an ldproxy convention - is to include the temporal resolution of the data in the Collection resource, either in the temporal extent object or as a top-level member. Something like `temporalResolution: DAY`. This could also be useful in general, but currently this is not covered in the standard. I thought we had a related issue somewhere, but I did not find it.

> Do you mean the `Add/✓` button or the `Apply` button at the top? I cannot reproduce the former, while the latter is the expected behavior since it triggers a page reload.

I meant the former. I tried it again and it sometimes happens (also from period to instant), but sometimes not. Here is a short animated GIF which shows how to tab switches from period to instant when pressing the `Add/✓` button.

period-instant

PZahnen:
@cportele - Thank you for the improvement ideas. I implemented the desired changes. I was not able to reproduce the unwanted switching between period and instant though. Maybe one of the changes I made in the meantime solved this issue.

cportele:
@PZahnen 

>  I was not able to reproduce the unwanted switching between period and instant though. Maybe one of the changes I made in the meantime solved this issue. 

I am also not able to reproduce the behavior with the current version.

rkoblenzer:
I've tested the time slider (with a non-public API). Generally it looks good and it works fine – with one exception: 

I can change date values manually, apply the changes and edit them again, and so on. But, after using the slider and applying with apply button once, no further manual changes are accepted: dates are reset to their previous values when I click on ✓ button after changing them manually.

I have also a remark that relates to filter editor in general: It would be nice to get a confirmation (in terms of a feedback, maybe an animation or a slightly different color like bright green) when values are changed with ✓ button (replaced by new values).

cportele (approved):
Everything I tested worked as expected.

azahnen (approved):

PR #938, opened 2023-03-15
Add option to suppress global CRS list

### LINKED ISSUE #937: Option to suppress the global CRS list
For APIs with a larger number of collections that support the same CRSs, there standard supports [a global CRS list in the Collections resource](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html#_global_list_of_crs_identifiers). If the CRS list of a collection includes a value `#/crs` then all CRSs in the global list are also supported for the collection. 

ldproxy currently always provides the CRSs defined on the API level in the global CRS list.

It is likely that some clients will not implement support for the global list (the current QGIS implementation is an example), so an option to suppress the global list would be useful to support such clients, too.

In version 4.0, it should be considered to not provide the global list as the default.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #937

### DIFF
--- a/ogcapi-stable/ogcapi-crs/src/main/java/de/ii/ogcapi/crs/domain/CrsConfiguration.java
+++ b/ogcapi-stable/ogcapi-crs/src/main/java/de/ii/ogcapi/crs/domain/CrsConfiguration.java
@@ -7,10 +7,10 @@
  */
 package de.ii.ogcapi.crs.domain;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import de.ii.ogcapi.foundation.domain.ExtensionConfiguration;
 import de.ii.xtraplatform.crs.domain.EpsgCrs;
-import de.ii.xtraplatform.docs.DocIgnore;
 import java.util.Set;
 import javax.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-938.surge.sh

ogcapi-stable/ogcapi-crs/src/main/java/de/ii/ogcapi/crs/domain/CrsConfiguration.java
```diff
@@ -56,6 +58,26 @@ abstract class Builder extends ExtensionConfiguration.Builder {}
...
+   * @default false
+   */
+  @Nullable
+  Boolean getSuppressGlobalCrsList();
+
+  @Value.Derived
+  @DocIgnore
+  default boolean shouldSuppressGlobalCrsList() {
+    return Boolean.TRUE.equals(getSuppressGlobalCrsList());
+  }
```
azahnen, line 79: It seems that `@DocIgnore` does not work correctly, this is visible in the docs. I will look at that, but in this case you should use `@JsonIgnore` anyway. 
`@DocIgnore` is only needed for very rare cases where an option should be read and/or written from/to json but should not show up in the docs.

azahnen (approved):
Works as expected besides the minor issue I commented on.

PR #944, opened 2023-03-21
Subdecoders and json schema

### LINKED ISSUE #907: Support JSON-values in SQL databases
Add support for JSON values in the columns of a table used in an SQL feature provider. This capability supports simpler database schemas and reduce joins.

In PostgreSQL the column could be of type `jsonb`, `json` or a string column. `jsonb` will usually be the best option, if the order of the members is not important for objects.

In SQLite the type would be `TEXT`. SQLite has features to support such columns.

The content would be defined in the feature schema as usual, i.e. the schema of the JSON content would be specified explicitly. 

An extension in the `sourcePath` to identify a column as JSON is probably necessary.

A simple example:

**SQL table**

| id | name | link | numbers |
| --- | --- | --- | --- |
| 1 | one | {"label": "foo", "url": "https://foo.bar"} | [ 1, 2, 3 ]

**Mapping**

```yaml
types:
  example:
    sourcePath: /example
    type: OBJECT
    properties:
      id:
        sourcePath: id
        type: INTEGER
        role: ID
      name:
        sourcePath: name
        type: STRING
      link:
        sourcePath: '[JSON]link'
        type: OBJECT
        objectType: Link
        properties:
          title:
            sourcePath: label
            type: STRING
          href:
            sourcePath: url
            type: STRING
      numbers:
        sourcePath: '[JSON]numbers'
        type: VALUE_ARRAY
        valueType: INTEGER
``` 

**GeoJSON instance**

```json
{
  "type": "Feature",
  "id": 1,
  "geometry": null,
  "properties": {
    "name": "one",
    "link": {
      "title": "foo",
      "href": "https://foo.bar"
    },
    "numbers": [1, 2, 3],
  }
}
```

Support for such properties or their sub-properties as queryables or sortables could be added in the future, but would not be required initially.

### LINKED ISSUE #908: JSON Schema as feature schema
An ldproxy feature schema is defined in YAML/JSON ([documentation](https://docs.ldproxy.net/providers/feature/#types)).

For cases where an external JSON Schema exists that is consistent with the schema in the feature provider, this JSON Schema could be used to generate the feature schema from the JSON Schema during start-up.

A new keyword `schema` would be defined to reference the JSON Schema document.

`schema` could be used for the complete schema of a feature type or for any sub-property.

Examples:

```yaml
types:
  foo:
    sourcePath: /foo
    schema: foo.json
  bar:
    sourcePath: /bar
    type: OBJECT
    properties:
      id:
        sourcePath: id
        type: INTEGER
        role: ID
      name:
        sourcePath: name
        type: STRING
      link:
        sourcePath: '[JSON]link'
        schema: link.json
```

Only a subset of JSON Schema would be supported, i.e., the JSON Schema must be directly mappable to an ldproxy feature schema.

### LINKED ISSUE #931: support allOf in provider schema
This will introduce the `allOf` composition from JSON schema to the provider schema. There are two known use cases:

**Define only some properties using an external JSON schema (#908)** 

```yaml
types:
  example:
    sourcePath: /main
    type: OBJECT
    allOf:
    - properties:
        id:
          sourcePath: id
          type: INTEGER
          role: ID
    - schema: names.json
```

This comes in really handy when used along with JSON columns (#907):

```yaml
types:
  example:
    sourcePath: /main
    type: OBJECT
    allOf:
    - properties:
        id:
          sourcePath: id
          type: INTEGER
          role: ID
    - sourcePath: '[JSON]names'
      schema: names.json
```

**Using columns from a joined table in the main feature**

This was already supported but pretty cumbersome since the join had to be repeated for every property.

*Before*

```yaml
types:
  example:
    sourcePath: /main
    type: OBJECT
    properties:
      id:
        sourcePath: id
        type: INTEGER
        role: ID
      name1:
        sourcePath: '[id=id]names/name1'
        type: STRING
      name2:
        sourcePath: '[id=id]names/name2'
        type: STRING
```

*After*

```yaml
types:
  example:
    sourcePath: /main
    type: OBJECT
    allOf:
    - properties:
        id:
          sourcePath: id
          type: INTEGER
          role: ID
    - sourcePath: '[id=id]names'
      properties:
        name1:
          sourcePath: name1
          type: STRING
        name2:
          sourcePath: name2
          type: STRING
```

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #907.
Closes #908.
Closes #931.

Depends on interactive-instruments/xtraplatform-spatial#195.

### DIFF
--- a/ldproxy-cfg/src/main/java/de/ii/ldproxy/cfg/EntityFactories.java
+++ b/ldproxy-cfg/src/main/java/de/ii/ldproxy/cfg/EntityFactories.java
@@ -46,7 +46,8 @@ public interface EntityFactories {
               public void addEntityGoneListener(Consumer<PersistentEntity> listener) {}
             })
         .add(
-            new FeatureProviderSqlFactory(null, null, null, null, null, null, null, null, null) {
+            new FeatureProviderSqlFactory(
+                null, null, null, null, null, null, null, null, null, null) {
               @Override
               public CompletableFuture<PersistentEntity> createInstance(EntityData entityData) {
                 return CompletableFuture.completedFuture(null);
--- a/ldproxy-cfg/src/main/java/de/ii/ldproxy/cfg/EntityFactories.java
+++ b/ldproxy-cfg/src/main/java/de/ii/ldproxy/cfg/EntityFactories.java
@@ -46,7 +46,7 @@ public interface EntityFactories {
               public void addEntityGoneListener(Consumer<PersistentEntity> listener) {}
             })
         .add(
-            new FeatureProviderSqlFactory(null, null, null, null, null, null, null, null) {
+            new FeatureProviderSqlFactory(null, null, null, null, null, null, null, null, null) {
               @Override
               public CompletableFuture<PersistentEntity> createInstance(EntityData entityData) {
                 return CompletableFuture.completedFuture(null);
--- a/ogcapi-draft/ogcapi-crud/src/main/java/de/ii/ogcapi/crud/app/EndpointCrud.java
+++ b/ogcapi-draft/ogcapi-crud/src/main/java/de/ii/ogcapi/crud/app/EndpointCrud.java
@@ -41,7 +41,7 @@ import de.ii.xtraplatform.crs.domain.CrsInfo;
 import de.ii.xtraplatform.crs.domain.EpsgCrs;
 import de.ii.xtraplatform.features.domain.FeatureProvider2;
 import de.ii.xtraplatform.features.domain.FeatureQuery;
-import de.ii.xtraplatform.features.domain.FeatureSchema.Scope;
+import de.ii.xtraplatform.features.domain.FeatureSchemaBase;
 import de.ii.xtraplatform.features.domain.ImmutableFeatureQuery;
 import de.ii.xtraplatform.features.domain.ImmutableFeatureQuery.Builder;
 import de.ii.xtraplatform.web.domain.ETag.Type;
@@ -380,7 +380,7 @@ public class EndpointCrud extends EndpointSubCollection implements ConformanceCl
                     .map(s -> s.substring(1, s.length() - 1))
                     .map(s -> EpsgCrs.fromString(s))
                     .orElseGet(coreConfiguration::getDefaultEpsgCrs))
-            .schemaScope(Scope.MUTATIONS)
+            .schemaScope(FeatureSchemaBase.Scope.MUTATIONS)
             .eTag(Type.STRONG);
 
     FeatureQuery eTagQuery =
@@ -448,7 +448,7 @@ public class EndpointCrud extends EndpointSubCollection implements ConformanceCl
                     .map(s -> s.substring(1, s.length() - 1))
                     .map(s -> EpsgCrs.fromString(s))
                     .orElseGet(coreConfiguration::getDefaultEpsgCrs))
-            .schemaScope(Scope.MUTATIONS)
+            .schemaScope(FeatureSchemaBase.Scope.MUTATIONS)
             .eTag(Type.STRONG);
 
     FeatureQuery eTagQuery =
--- a/ogcapi-draft/ogcapi-crud/src/main/java/de/ii/ogcapi/crud/app/QueryParameterSchemaFeatures.java
+++ b/ogcapi-draft/ogcapi-crud/src/main/java/de/ii/ogcapi/crud/app/QueryParameterSchemaFeatures.java
@@ -17,7 +17,7 @@ import de.ii.ogcapi.foundation.domain.HttpMethods;
 import de.ii.ogcapi.foundation.domain.OgcApiDataV2;
 import de.ii.ogcapi.foundation.domain.OgcApiQueryParameter;
 import de.ii.ogcapi.foundation.domain.SchemaValidator;
-import de.ii.xtraplatform.features.domain.FeatureSchema.Scope;
+import de.ii.xtraplatform.features.domain.FeatureSchemaBase;
 import de.ii.xtraplatform.features.domain.ImmutableFeatureQuery;
 import io.swagger.v3.oas.models.media.Schema;
 import java.util.Map;
@@ -104,7 +104,7 @@ public class QueryParameterSchemaFeatures extends ApiExtensionCache
     if (parameters.containsKey(getName())
         && Objects.equals(parameters.get(getName()), "receivables")) {
       try {
-        queryBuilder.schemaScope(Scope.MUTATIONS);
+        queryBuilder.schemaScope(FeatureSchemaBase.Scope.MUTATIONS);
       } catch (NumberFormatException e) {
         // ignore
       }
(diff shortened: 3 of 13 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-944.surge.sh

cportele:
@azahnen 

I have problems to get started with testing. Here is a DB table that I am using:

```sql
CREATE TABLE public.features (
	"_id" serial4 NOT NULL,
	id text NULL,
	properties jsonb NULL,
	geom public.geometry NULL,
	fid int4 NULL,
	featuretype text NULL,
	CONSTRAINT features_pkey PRIMARY KEY (_id)
);
```

I tried `auto` mode, which was not successful: the `jsonb` column was ignored and there was an exception that the FeatureSchema could not be created because `name` was missing. However, a schema was created, but an incomplete one.

I have then updated to schema as follows:

```yaml
types:
  features:
    sourcePath: /features
    type: OBJECT
    allOf:
    - properties:
        _id:
          sourcePath: _id
          type: INTEGER
          role: ID
        id:
          sourcePath: id
          type: STRING
        featuretype:
          sourcePath: featuretype
          type: STRING
        fid:
          sourcePath: fid
          type: INTEGER
        geom:
          sourcePath: geom
          type: GEOMETRY
          role: PRIMARY_GEOMETRY
          geometryType: ANY
    - sourcePath: '[JSON]properties'
      schema: https://www.portele.de/xplanung/pfs_properties.json
```

I can access the features, but `properties` is ignored. I did have the time to debug the reasons. Do you have an idea what I am doing wrong? Thanks!

I used log level DEBUG, but there were no messages related to problems in the log.

azahnen:
@cportele I started to implement unit tests for the schema resolver and looked into https://www.portele.de/xplanung/pfs_properties.json to see if I could add that. The top-level `oneOf` is problematic since there is no way to select an element, it will always take the first one. If the `oneOf` would consist of `$ref` elements, you could reference the according `$defs` in `schema`.

Do you still work with that schema or do you have a customized variant somewhere?

azahnen:
@cportele The `oneOf` would work the same but would not play a role anymore since you can directly reference `$defs`. Since I assumed that there is one `oneOf` element per feature type, you could then do something like this: 
```yaml
schema: https://www.portele.de/xplanung/pfs_properties.json#/$defs/PFS_Schutzgebiet
```

azahnen:
@cportele I was able to reproduce the issue with nested objects (`Invalid sourcePath in provider configuration`), I pushed a fix.

cportele:
@azahnen - Yes, I am testing it now, this seems to fix this exception. Thanks.

cportele:
@azahnen - I have created a `jsontest` repo with the configuration that I am using for testing.

Here are some issues that I have noticed so far: 

- [ ] For the first feature (in each request) the properties are as expected. Starting with the second feature, the properties from the JSON column are embedded in an `_IGNORE_` member.
- [ ] `SchemaDeriverOpenApi.withName()` throws an exception, because `oapiSchema` is `null` (the schema is identified as a value, but is not a value and therefore stays `null`).
- [ ] The order of the properties does not follow the order in the JSON Schema. I guess that is not really an issue, more an observation since we probably cannot control how a) the properties in the `properties` member are parsed when reading the schema and b) for a `jsonb` column the order of members in a object is not predictable. Or is there a way to process the schema so that the order in the feature schema is the same as in the JSON Schema? Then using a `json` column should maintain the order?

Could you have a look at these issues?

azahnen:
@cportele 
- The first two issues should be fixed. 
- I could not reproduce the third issue, even without the recent fixes. How can I trigger the exception?
- Regarding the sort order, order from `json` columns should be maintained now. The sorting transformer falsely reordered some json properties, it should ignore them for now. 
I also looked into making the sorting transformer more generic so that it could reorder the properties for `jsonb` columns, but that would be quite complex. Currently it only handles a very specific case where values from a row are intercepted by a joined query. Handling any order and missing properties would require a completely different mechanism.
Note that all of this has nothing to do with the schema, the feature schema retains the order from the JSON Schema.

cportele:
@azahnen 

Yes, the first two issues are resolved. 

From the description it also sounds like we should consider the ordering issue resolved - I will do some tests over the weekend with `json` columns. This is mainly relevant, if there is a format where the order is important (GML).

I also no longer see the exception. It was still there yesterday (a type of a property was `UNKNOWN`), but maybe that was before the changes to the JSON Schema. I will also do more tests with schema variations.

Thanks!

cportele:
@azahnen - I have made more tests and I had to make some code changes to get the correct/better results. Since some of that may not be the right or best way, I have done this in separate branches. Change overview:

ldproxy (`subdecoders-and-json-schema-2`):

* remove `_IGNORE_` in OpenAPI Schemas
* remove `_IGNORE_` in JSON Schemas of returnables
* remove `_IGNORE_` in the values of the `properties` parameter

xtraplatform-spatial (`subdecoders-2`):

* `[JSON]column_name` did not work for a property. This was mapped to a column called `JSON`.
* JSON values that are, e.g., just a number were not parsed, the token was `NOT_AVAILABLE`. The proper approach as I understand it would be to send `endOfInput()` to the feeder, but this conflicts with the way the parser is used. For now I have added a whitespace character at the end so that the parser can provide the correct token, but that is a hack.
* I have added warnings when `allOf`/`anyOf`/`oneOf` are encountered as well as code to process constraints that can be mapped to `SchemaConstraints`.

There several TODOs and unfinished new code (e.g. `GEO_DEFS` only has polygon geometries). Before merging into master, we should clean-up the new code.

In addition, we should add more documentation about the expectations/limitations of `schema` in https://interactive-instruments-ldproxy-pr-944.surge.sh/de/providers/feature/#types.

cportele:
@azahnen 

I did tests and found more issues: 

* `_IGNORE_` is also included in the property names after flattening, e.g., in CSV. This would require changes to `WithTransformationsApplied()`. Instead of adapting more existing (and future) code with treatment for this special case it would be better to find another approach so that we can avoid code changes that will be difficult to maintain.
* Flattening values from JSON columns also seems to require special treatment. A JSON value `[ 1.2, 2.4 ]` of a property `test` (`type: VALUE_ARRAY`, `valueType: FLOAT`) is flattened to two properties `test.test`, which in turn results in an map key exception because there are two properties/keys with the same name.

azahnen:
@cportele

**json colums**
- I added support for values and value arrays.

**allOf**
- I replaced the `_IGNORE_` logic with a better aproach as discussed.

These two changes should fix all issues regarding flattening and schemas.

**external schema**
- I added support for reusable schema `fragments`. By default only entries from `fragments` can be referenced in `schema`.
- I moved the JSON schema resolver to an extension that has to be enabled to use JSON schemas in `schema`. That makes it easier to add configuration options and also adds a separate page in the documentation.
- I added a configuration option to map `$defs` to `objectType`s, so you do not have to define all `Link`s manually.
- I merged the constraints derivation from your branch.
- I did not merge the warnings. I think warnings that the user cannot fix are just annoying. It should be enough that the limitations are described in the documentation.
- I updated the `jsontest` repo to make use of the improvements.

cportele:
@azahnen - I checked all issues and they are all gone now. I will do more testing tomorrow. 

`fragments` is helpful and to make `JSON_SCHEMA` an extension looks like the right approach to me.

> I did not merge the warnings. I think warnings that the user cannot fix are just annoying. It should be enough that the limitations are described in the documentation.

This is something that we should discuss. I am more annoyed by the silent re-interpretation of the schema (and I did adapt/fix the schema/configuration, once I found out that the oneOf was ignored).

azahnen:
@cportele I could imagine logging the warnings when either `typeValidation` or a new flag in `JSON_SCHEMA` is enabled. But there should be some way to get rid of the warnings besides setting the log level to `ERROR`.

cportele:
@azahnen - Moving the warning checks to `typeValidation` during start-up would probably be best. I will have a look.

azahnen:
@cportele They can stay in the same place in `JsonSchemaResolver.resolveComposition`. You just have to pass `data` as parameter and check for `typeValidation != NONE`.

cportele (approved):

PR #946, opened 2023-03-31
Add option to suppress empty collections in the HTML page for the Feature Collections

### LINKED ISSUE #945: Add option to suppress empty collections in the HTML page for the Feature Collections overview
For cases where the feature provider contains a larger number of types and only a subset of them have items in a dataset, it can be helpful to users browsing the dataset to only show the collections with data. 

Add an option to suppress collections in the HTML page of the Feature Collections resource where the known item count is 0. If the item count is unknown the collection should be included.

Note that this would only affect the HTML representation of the Feature Collections resource. The collection should still be a resource in the API.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #945 .


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-946.surge.sh

azahnen (approved):
LGTM

PR #954, opened 2023-04-26
Redesign tiles2

### LINKED ISSUE #845: tiles follow-up
- [x] ~TileMatrixSets should be entities, not resources~ moved to #877
- [x] publish metadata like zoomLevels + center in `TileProvider`, use in API (see `TilesDataHydrator.process`)
- [x] ~~integrate `ogcapi-maps` into `ogcapi-tiles`~~ delete `ogcapi-maps`

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #845.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-954.surge.sh

cportele:
Note to self: I received errors when `layers` / `layerDefaults` was still used in the tile provider configurations. After changing to `tilesets` / `tilesetDefaults` the APIs started. To check why the alias did not work.

cportele:
@azahnen - I have a problem getting zoomstack to work. If I change the configuration to the following:

Service:

```yaml
- buildingBlock: TILES
  enabled: true
```

Provider:

```yaml
---
id: zoomstack-tiles
providerType: TILE
providerSubType: MBTILES
tilesets:
  __all__:
    id: __all__
    source: OS_Open_Zoomstack.mbtiles
```

This results in a Features tile provider, not an MBTiles tile provider, which always returns empty tiles.

I then tried an explicit tile provider identifier:

Service:

```yaml
- buildingBlock: TILES
  enabled: true
  tileProviderId: zoomstack-mvt
```

Provider:

```yaml
---
id: zoomstack-mvt
providerType: TILE
providerSubType: MBTILES
tilesets:
  __all__:
    id: __all__
    source: OS_Open_Zoomstack.mbtiles
```

This results in "Tile provider with id 'zoomstack-mvt' not found."

It does not seem to load the zoomstack tile provider. The earthatnight tile provider, which is also a MBTiles provider is loaded though.

Any idea what I am missing / doing wrong?

azahnen:
@cportele  There were multiple issues that should be fixed now:

- When the path to an MBTiles file did not exist, the tile provider failed silently since the exception was thrown in the constructor. It will now log an error.
- When no running provider was found and `tileProviderId` was not set explicitly, the API executed the auto migration. It created a features tile provider in your case since that is the default. It will now also check for failed providers before executing the auto migration.
- The source paths were resolved against the data folder instead of the tiles folder. The correct path should now be `zoomstack/OS_Open_Zoomstack.mbtiles`.

cportele:
@azahnen - Thanks, all the changes work, maybe except:

> The source paths were resolved against the data folder instead of the tiles folder. The correct path should now be zoomstack/OS_Open_Zoomstack.mbtiles.

When I tried it, ldproxy expected the file at `api-resources/tiles/OS_Open_Zoomstack.mbtiles`.

azahnen:
@cportele What I tried to say is, you should set `source` to `zoomstack/OS_Open_Zoomstack.mbtiles` if the file resides in `api-resources/tiles/zoomstack/OS_Open_Zoomstack.mbtiles`.

cportele:
Ah ok, got it 😄

azahnen:
@cportele I did test the HTTP tile provider, but not with MAP_TILES. The encodings have to be configured in the tile provider. I just updated the docs (wait for the build), it contains an example.

cportele:
I see the `encoding` property in the recent commit. Thanks!

cportele:
@azahnen 

> I did test the HTTP tile provider

It does not work for me, also with the example configuration. There are two issues:

`levels` is ignored when deserializing the HTTP tile provider, both in `tilesetDefaults` and in `tilesets`. As a result, there is no tileset (no tile matrix set).

For `tilesetDefaults`, if I remove the default for `getTilesetDefaults()` and change it to the following, `levels` are read correctly (but I get an error, if no defaults are specified):

```java
@Override
TilesetHttpDefaults getTilesetDefaults();
```

For `tilesets` this seems to be more complex.

Since it works for the Features tile provider, I started to update the HTTP tile provider along the lines of that tile provider, but stopped again, because it is probably not necessary, since it works for you.

The second issue is in `TileStoreHttp.getUrl()`. This uses `StringTemplateFilters`, which does not work as it expects templates to be wrapped in double curly brackets, but URI templates use a single curly bracket. I have to replace it with the following to make it work:

```java
  private static String getUrl(TileQuery tile, String template) {
    return template
        .replace("{layer}", tile.getTileset())
        .replace("{tileMatrixSet}", tile.getTileMatrixSet().getId())
        .replace("{tileMatrix}", String.valueOf(tile.getLevel()))
        .replace("{tileRow}", String.valueOf(tile.getRow()))
        .replace("{tileCol}", String.valueOf(tile.getCol()))
        .replace("{fileExtension}", EXTENSIONS.get(tile.getMediaType()));
  }
```

Since the example configuration works for you, I must be doing something wrong.

azahnen:
@cportele 
The `urlTemplate` issue is a copy and paste error in the example, it is correct in the `urlTemplate` docs description, so you should use double curly brackets. I will fix the example.

Regarding `levels`, it is not working anymore for features either. I made some changes for the docs yesterday which must have broken it, I will look into it.

cportele:
@azahnen - OK. However, the `urlTemplate` should use single curly brackets, not double curly brackets. We should change the implementation, not the example.

azahnen:
@cportele Why? I changed that intentionally since we are using double curly brackets for substitutions everywhere else in the configuration and I found it confusing.

cportele:
We use double curly braces where this is driven by mustache/JavaScript. 

On the other hand, [URI templates](https://www.rfc-editor.org/rfc/rfc6570) use single curly braces, so we should use single curly braces whenever the value is a URI template.

azahnen:
@cportele I was not talking about mustache, but configuration options like `stringFormat`, `featureTitleTemplate` and `labelTemplate`. As a user I would expect that to be consistent.

I pushed a fix for the levels issue.

cportele:
Yes, we use double curly braces as a ldproxy convention for [string templates with additional filters](https://docs.ldproxy.net/de/providers/details/transformations.html#string-template-filter). 

A URI template is something different. It is specified by [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570) and used in the OGC API standards. As an API author I would expect that the value for a Tile URI template is consistent with those standards. We could debate whether we use names typically used in the OGC documents (`tileMatrix`, `tileRow`, `tileCol`) or the ones expected in other tools (`z`, `y`, `x`). Since ldproxy is an OGC API implementation using the former seems reasonable to me.

We can of course also define a configuration property for the tile URIs that supports the ldproxy string templates instead of a URI template. In that case this should be clear(er) from the property name (e.g., `tileUrlStringTemplate`) and not imply that the value is a URI template. That would make sense, if there would be a need to use string template filters in configurations.

Let's discuss that in our next meeting.

cportele:
@azahnen - I made some changes after more testing with a updated tiles implementation. Please have a look at my commits in xtraplatform-spatial and ldproxxy.

As far as I can see, the following topics need a discussion before merging:

- I would update the use of `layer` to `tileset` in many places, currently the code is hard to understand unless one is aware of the change in terminology. I can make the changes.
- We should cleanup code, in particular code that has been commented out. And review the remaining TODOs, if any of them need to be addressed. 
- Discuss the Map Tiles module. I would actually remove the module as it currently adds no value:
  - Currently it does not properly work with the tile providers and needs updates.
  - However, since the current design only supports a single tile provider per API, having Map Tiles in addition to Tiles does not make sense without significant changes. So for now, if one wants to add image tiles in additional to vector tiles generated from a feature provider, this requires separate APIs anyhow.
  - Since I can also define an HTTP or MBTiles tile provider in the Tiles module, there is no point in having the Map Tiles module at the moment. The only option that it adds is that the path is not `.../tiles/...` but `.../map/tiles/...`.
  - We want to get rid of the module anyhow. 
- Semantics of `urlTemplate` as discussed aabove.

cportele:
The updated demo configurations for the changes are in https://github.com/ldproxy/demo/tree/redesign-tiles2

cportele (approved):
I have completed my tests and edits.

There are a number of TODOs that are not clear to me. I have marked them with `(???)`. Could you review them and see, if they need any action now, if they are no longer relevant and the TODO can be removed or if we can address them later (in that case provide more info what should be changed in the future. Once the `(???)` markers are resolved, the PR can be merged.

I worked on setting additional metadata in the MBTiles caches, but this is getting too complicated to finish now, so I left those as TODO for later.

cportele:
@azahnen - Wait, my last commit fails with an error. I need to fix that first. Stay tuned...

cportele:
@azahnen 

Here is the relevant excerpt from the daraa-tiles configuration:

```yaml
...
tilesetDefaults:
  center:
    lon: 36.1033
    lat: 32.6264
  levels:
    WebMercatorQuad:
      min: 6
      max: 18
      default: 13
    WorldCRS84Quad:
      min: 5
      max: 17
      default: 12
    WorldMercatorWGS84Quad:
      min: 6
      max: 18
      default: 13
...
```

The result is:

> Cannot build LonLat, some of required attributes are not set [lon, lat]

azahnen:
@cportele I fixed the 'center' issue.

PR #955, opened 2023-04-26
GeoPackage upload in manager

### LINKED ISSUE #504: support GPKG in the Manager

PZahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #504 .

Depends on interactive-instruments/xtraplatform#130.

Only merge after #956.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-955.surge.sh

cportele (approved):
Tested with two GeoPackages and (almost) worked as expected.

The "almost" is not related to the upload in the manager, but to the GPKG auto mode. I uploaded the daraa.gpkg file and the features all are in Turkey. The CRS is detected as EPSG:4326 with no force on the axis order. This is not correct, GPKG always uses LON_LAT (or Easting/Northing, etc.). This is the statement in the [standard](http://www.geopackage.org/spec131/index.html#gpb_format):

> The axis order in WKB stored in a GeoPackage follows the de facto standard for axis order in WKB and is therefore always (x,y{,z}{,m}) where x is easting or longitude, y is northing or latitude, z is optional elevation, and m is optional measure. This ordering explicitly overrides the axis order as specified in the SRS metadata, applying Case 4 from OGC 08-038r7, Revision to Axis Order Policy and Recommendations. This was done to maintain consistency with previous implementations of WKB that predated the OGC policy.

While it is easy to update this manually in the provider yaml, It would be good to handle this correctly when using the manager.

azahnen:
@cportele Could you create a separate issue for that? As you say this is unrelated to the PR and I am not sure I will be able to look into that before the release.

cportele:
Done: https://github.com/interactive-instruments/ldproxy/issues/959

cportele:
Another comment or more an open question: When deleting the newly created services again, the provider/service configurations are deleted, but not the GPKG file. Should it be deleted (when it was uploaded using the manager)?

azahnen:
@cportele I think that is out of scope for now. I would like to refocus the manager as an authoring tool around the new store sources concept, which would then also include resources/blobs handling, but we should discuss that somewhere else.

cportele:
@azahnen - Yes, that makes sense.

PR #956, opened 2023-05-01
use blob store instead of fs

### LINKED ISSUE #877: use store for all file-like inputs and outputs
To enable usage of remote store sources like S3, any direct filesystem access has to be removed. The `xtraplatform-store` module provides abstractions that should cover all cases.

The only exception are temporary files that are not expected to persist.

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #877.

The review goal would be that everything works like before.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-956.surge.sh

cportele (approved):
I have tested

- features (pgis and gpkg providers)
- resources (read/write)
- styles (read/write)
- tiles (providers: features/mbtiles; cache: plain/mbtiles)
- tiles3d

without any problems.  I still have to test routes, jsonld-contexts and queries, but based on the results so far I can also do this in the pre-release tests.

An open issue is documentation, but I guess this can be added when there are other stores than the file stores.

Edited to add that I did my tests with my redesign-tiles2 changes merged into the blob-store branches.

PR #960, opened 2023-05-03
Edits from pre-release tests

cportele:
### Pull request checklist

-   [ ] Updated documentation

- Fix regression: Title and description were not set for the tilesets in the tilesets metadata. In addition: removed unused vars and simplified code.
- improve featureType->collectionId mapping
- no ETag for HTML unless explicitly configured: This issue was probably already present in 3.3 as a result of the CRUD changes.
- Tiles: add default for center. Regression tests showed that previously there was a center value if a default level was specified, but no center - although I could not really find easily how/where that was computed. Nevertheless it seems reasonable to use the center of the extent as a fallback.
- IS NULL not supported by WFS provider. IS NULL is required to properly implement the `bbox` and `datetime` query parameters, which match all features where the primary geometry or instant is `null`. This has been corrected as part of the recent refactoring of the FeaturesQueryImpl class. For WFS we simply continue to ignore the condition, i.e. drop the IS NULL predicate.
- See also https://github.com/interactive-instruments/xtraplatform-spatial/pull/202.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-960.surge.sh

PR #975, opened 2023-05-20
Schema refs

### LINKED ISSUE #970: Feature Schema: Data type for references to features

So far, references to other features are typically encoded as a `Link` object with `href` and `title` properties. In some cases, it would better to have use the "foreign key" (the `featureId` of the referenced feature) instead.

For referenced features in the same API we also need the `collectionId`. For features in another API implementing OGC API Features we also need the URI of the `apiLandingPage`.

NOTE: The `Link` object approach can still be used. Existing configurations will continue to work without changes.

### Describe the Use Case

An example where the "foreign key" would be better:

A client has a feature collection where the features reference features in another collection. The clients wants to fetch the related features to process their information.

In the case of links, the client has to sent one request per link (or try to parse the link URIs for the feature ids).

In the case of foreign keys, the client can send a single request to retrieve the features, e.g. using a CQL2 filter like `id IN (1, 3, 7, 16, 53, 59, 64, 99)` where `id` is a queryable representing the feature id.

Add new types `FEATURE_REF` and `FEATURE_REF_ARRAY` to the feature schema. The value type of the property can be `STRING` or `INTEGER`, depending on type of the feature property with the role `ID`. `STRING` is the default.

To indicate the target feature type, a new annotation `refType` can be used. Note that since the schema does not know anything about OGC API or Collections, this is the id of the feature type in the schema, not the id of the collection in the API. When an API link is derived for the reference, the feature type id may need to be mapped to the collection id.

```yaml
      abs:
        label: Abschnitt/Ast
        description: 16-stellige Kennung des Abschnittes oder Astes
        type: FEATURE_REF
        refType: abschnitteaeste
        sourcePath: abs
```

For features that are not published in the same API, instead of `refType` we could provide a URI template in a new annotation `refUriTemplate` with `{value}` for the foreign key so that a link can be constructed when needed. Both single and double curly braces should be supported.

```yaml
      abs:
        label: Abschnitt/Ast
        description: 16-stellige Kennung des Abschnittes oder Astes
        type: FEATURE_REF 
        refUriTemplate: 'https://demo.ldproxy.net/strassen/collections/abschnitteaeste/items/{value}'
        sourcePath: abs
```

In addition, the schema of the features in a collection should be adapted as discussed in a [proposal in the OGC Features API working group](https://github.com/opengeospatial/ogcapi-features/issues/740#issuecomment-1501758135).

### Additional Information

_No response_

### LINKED ISSUE #972: Controlling the encoding of feature relationships

This builds on #970 which introduces feature references. As described in that issue, clients will prefer different representations depending on the situation. 

### Describe the Use Case

Different clients / use cases will prefer different representations.

For example, for presentation to a human, a title and a resolvable URI may be important. 

For cases where the client has knowledge about the data structures and wants to process data, just the local id of the referenced object will often be easier and sufficient. 

See also #970. 

Note: most of the text is taken from a proposal under discussion in the OGC Features API working group.

It should be possible to request the encoding of properties that are of type `FEATURE_REF` or `FEATURE_REF_ARRAY` depending on the need of the client. That is, without changing the media type of the response, clients should be able to request the encoding of feature references.

### Profiles

IETF has specified [RFC 6906](https://www.rfc-editor.org/rfc/rfc6906) for such cases using the concept of profiles:

> A profile is defined not to alter the semantics of the resource representation itself, but to allow clients to learn about additional semantics (constraints, conventions, extensions) that are associated with the resource representation, in addition to those defined by the media type and possibly other mechanisms.

The idea would be to define profiles for different ways to express a relationship in an instance. For now, we should implement the three profiles described below.

The examples below use a GeoJSON representation, but this could apply to other media types, too. 

Media types may and often will only support a subset of the profiles. An example would be media types that only support scalar property values.

### Profile `rel-as-link`

As a link object with a URI and a human readable title:

```json
{
    "type": "Feature",
    "id": 1,
    "geometry": {...},
    "properties": {
        "type": "Primary",     
        "level": 1,
        "inMunicipality": {
            "href": "/collections/municipalities/items/15",
            "title": "Tanelorn"
        },
        "maintainers": [ {
            "href": "https://www.example.com/api/v1/road-maintenance/collections/organization/items/Acme",
            "title": "Acme Inc."
        } ]
    },
    "links:": [
        {
            "href": "http://www.opengis.net/def/profile/ogc/0/rel-as-link",
            "rel": "profile"
        }
    ]
}
```

### Profile `rel-as-uri`

As a URI:

```json
{
    "type": "Feature",
    "id": 1,
    "geometry": {...},
    "properties": {
        "type": "Primary",     
        "level": 1,
        "inMunicipality": "/collections/municipalities/items/15",
        "maintainers": [ "https://www.example.com/api/v1/road-maintenance/collections/organization/items/Acme" ]
    },
    "links:": [
        {
            "href": "http://www.opengis.net/def/profile/ogc/0/rel-as-uri",
            "rel": "profile"
        }
    ]
}
```

### Profile `rel-as-key`

Just the identifier (the profile consistent with the encoding-independent schema):

```json
{
    "type": "Feature",
    "id": 1,
    "geometry": {...},
    "properties": {
        "type": "Primary",     
        "level": 1,
        "inMunicipality": 15,
        "maintainers": [ "Acme" ]
    }
}
```

### Requesting a profile in a Features request

When requesting features via an API, a `profile` query parameter can be provided, e.g.: `GET /collections/foo/items?profile=rel-as-uri`. 

`rel-as-link` is the default profile. 

If a Feature encoding does not support a requested profile (or the default profile, if no profile was requested), no error is thrown, but a fallback profile is selected (clients can determine the selected profile from the Link headers in the response).

### Additional Information

This approach is not able to support one capability of the current approach to explicitly specify the link objects in the feature schema. In the explicit schemas, the `title` and `href` properties could be sourced from different source paths. In the `FEATURE_REF` case, this is always the feature id of the referenced feature. It is not possible to select a different, more descriptive property of the feature (e.g., a name property). 

For data where the feature ids are not useful enough for human users and useful titles are considered important, the approach with explicit link objects can still be used.

### LINKED ISSUE #973: Schemas: Use encoding-agnostic schema instead of describing the GeoJSON representation

The following was discussed in the OGC Features API working group:

> Instead of requirements to create a JSON Schema for a GeoJSON feature, we should provide a encoding-independent JSON Schema that specifies all the feature properties. This follows the same approach as the Queryables, Sortables and Tile Set Metadata resources.
> 
> Note that APIs that want to encode `describedby` links to a JSON Schema for validating GeoJSON or JSON-FG instances can include such links, but this would be orthogonal and not relevant for Part 4. The encoding-specific schemas can be derived from the encoding-independent JSON Schema.
>
> The encoding-independent JSON Schema should follow the [schema recommendations for the Queryables resource](https://docs.ogc.org/DRAFTS/19-079r1.html#rec_filter_queryables-schema).
> 
> The JSON Schema will use `readOnly` and `writeOnly` to identify properties that are only including in a feature in a response from the API (`readOnly`) or that are only relevant for CREATE/REPLACE/UPDATE operations (`writeOnly`).
> 
> The property that represents the `featureId` will be marked as `readOnly`. We could use a new keyword to tag that property as the local identifier of the feature (e.g., `x-ogc-role: ID`). Note that it is very likely that there will be [rules for custom keywords in JSON Schema](https://github.com/orgs/json-schema-org/discussions/329). That is, the keyword will depend on the result of that discussion.

This approach has several advantages, in particular it should no longer be necessary to provide different schemas for the returnables or the create, replace, or update operations. 

It is also related to the introduction of the `FEATURE_REF` type (#970).

### Describe the Use Case

A single schema to describe features, independent of the encodings that are supported by the API. 

See the description above.

### Additional Information

Consequences:

- Since schema derivation is no longer GeoJSON-specific, the schema derivation should be moved from GeoJSON to Features Core.
- The role information is not only added for the identifier and reference properties, but also for all other roles. Clients will also need to know, for example, which property is the primary geometry, if there are multiple geometry properties. Roles are using lowercase-hyphen formatting.
- The annotations are not only useful for the content schema, but also for the queryables and sortables. The annotations are added there, too.
- Through the use of an encoding-agnostic schema and the use of `writeOnly`/`readOnly` to identify properties that are relevant for CRUD operations, the various schema types are no longer needed.
- The collection schema is also no longer needed.
- Since there is only a single schema for a collection, it can be made available at `/collections/{collectionId}/schema`.
- We can rename "returnables" to "features" since the other types are gone.
- Support JSON Schema 2020-12 (which is now better supported) and make it the default.
- Use content negotiation to select the JSON Schema version (using draft-ietf-httpapi-rest-api-mediatypes-03).

cportele:
### Pull request checklist

-   [x] Updated documentation

- Closes #970 
- Closes #972 
- Closes #973

There are some small changes to the original design, I will update the issues.

This PR is still a draft, because there are two open issues:

- [ ] `rel-as-link` does not yet work, e.g., for GeoJSON. I tried to add this capability in `FeatureTokenTransformerSchemaMappings`, but was not successful. The current flow is difficult to understand and it is hard to know what side effects changes may have on other configurations.
- [ ] In `rel-as-link` it should be possible to "inject" a title from different source path(s), otherwise this option will not be useful for cases where the IDs have not thematic meaning (e.g., for auto-increment ids).


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-975.surge.sh

cportele:
For testing purposes, I have updated https://demo.ldproxy.net/strassen with the changes. Note that the GeoJSON is flattened.

The encoding of the different profiles for references (property `unfaelle.abs`):

* https://demo.ldproxy.net/strassen/collections/unfaelle/items?f=json
* https://demo.ldproxy.net/strassen/collections/unfaelle/items?f=json&profile=rel-as-link (the default, identical to the previous one)
* https://demo.ldproxy.net/strassen/collections/unfaelle/items?f=json&profile=rel-as-key (only the foreign key)
* https://demo.ldproxy.net/strassen/collections/unfaelle/items?f=json&profile=rel-as-url (only the URI)

This also works in the other formats (JSON-FG, HTML, CSV, FlatGeobuf), although CSV and FlatGeobuf only support rel-as-key and rel-as-uri for now.

Note: The selected profile is returned as a `Link` header with `rel=profile`.

The updated JSON schema for the accidents: https://demo.ldproxy.net/strassen/collections/unfaelle/schema

Notes:

* The `x-ogc-role` annotations identify special roles of the properties. That is, `fid` is the `featureId`; `unfzeit` the property taken e.g. in JSON-FG for `time`; `geometry` the property taken e.g. in GeoJSON/JSON-FG for `geometry` (resp. `place`). `abs` is a feature reference and the value is a `featureId` from the `abschnitteaeste` collection.
* `fid` is `readOnly` because it is an auto-increment column and cannot be set explicitly. Since it is also the `featureId` column, a new feature can only be created with POST on `/collections/unfaelle/items`. The `featureId` is assigned automatically.

azahnen:
@cportele I guess this depends on the branch in `xtraplatform-spatial`? Would you create a PR there as well?

azahnen (approved):
Looks good to me.

One question regarding the default profile, will that be defined in the spec? I guess otherwise we should make it configurable.

cportele:
The default profile still needs to be discussed in OGC. To explain, why I ended up with the current approach: 

My original idea was to use `rel-as-key` as the default, because this reflects the provider source (for an SQL backend). However, this does not work well for  the HTML representation where `rel-as-link` is probably almost always what one wants. And it reflects what we have used in most APIs for GeoJSON, too. The options that I saw:

- Pick `rel-as-link` as the default, but allow formats to not implement that profile.
- Do not define a global default, but allow that the default varies by format.

Eventually I opted for the first option, mainly because that is clearer in the OpenAPI definition (`profile` has a documented default), but thinking again about this I am not so sure, which is the better approach.

I would not make the (global) default configurable, mainly because of HTML. We could leave it as it is for now and wait for the discussion in OGC or we could switch to the second option. In practice, there is not really a difference. If we would use the second option, the defaults that I would see:

- HTML: `rel-as-link`
- GeoJSON/JSON-FG: `rel-as-link`
- GML: `rel-as-link` (`rel-as-key` would not be valid with most GML app schemas since an xlink is expected)
- CSV: `rel-as-key` (currently we don't support `rel-as-link`, so `rel-as-key` is the effective default there)
- FlatGeobuf: `rel-as-key` (same as with CSV)
- glTF: `rel-as-key` (same as with CSV)
- CityJSON: ? (the spec only specifies parent-child relations, I don't think there is a concept of relationships between different first-level objects)

So, I would leave it as it is for now a try to clarify it in OGC over the next weeks so that we can align the implementation before v3.5, if necessary. What do you think?

azahnen:
Yes, sounds good. That was more out of interest and for future planning, not something I would change in this PR.

Issue #846, opened 2022-12-16
roll out new query parameter handling


azahnen:
In #786 a new mechanism was introduced to move the evaluation of query parameters into the `QueryParameter` classes. This is already used for all query parameters that are used in `ogcapi-tiles`, it should be used everywhere.

PR #992, opened 2023-06-12
Support default values for typed query parameters

cportele:
### Pull request checklist

-   [ ] Updated documentation

Query parameters can have default values. Some `TypedQueryParameter` classes implemented the `parse()` method already to return the default value, if no value was provided with the request. However, most classes did not handle a `null` value and `parse()` was not executed for query parameters without a value. The result was that default values had to be taken care of separately.

This has been addressed:

- `QueryParameterSet.evaluate()` now also calls `parse()` for query parameters without a value.
- `parse()` now handles `null` values for all `TypedQueryParameter` classes.

This change is part of #846.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-992.surge.sh

azahnen (approved):
LGTM

PR #994, opened 2023-06-12
CRUD improvements

cportele:
### Pull request checklist

-   [x] Updated documentation

- [x] If the feature includes a property with the timestamp of the last update, this property can be tagged as `isLastModified: true` in the provider schema.
- [x] Add support for `Last-Modified`-based preconditions for CRUD operations on features. The value is added to `FeatureStream.Result` in addition to the ETag. 
- [x] The `Last-Modified` value and the spatial / temporal extents are derived in a new token transformer when fetching features. For reduced results (at the moment: for a single feature), the spatial and temporal extents are provided in HTTP headers `Content-Bounding-Box` and `Content-Temporal-Extent`. This is necessary pass the information from the Features Core handler to the CRUD handler, but could in general be useful for other clients, too.
- [x] Add the new HTTP headers to the API definition.
- [x] Adjust feature changes to support also the current spatial/temporal extents (for PUT/PATCH/DELETE). 
- [x] For DELETE, also fetch the current feature and evaluate preconditions.
- [x] Resolve TODO to use the `FeaturesQuery` class to generate the feature query for fetching the current feature. 

This pull request depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/211.

Follow-up tasks:

- Support precondition checks also for other CRUD operations (styles, style metadata, file resources, stored queries).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## v3.4.2 (08/08/2023)
+
+#### Fixed bugs
+
+-  fix feature mutation operations [#1027](https://github.com/interactive-instruments/ldproxy/issues/1027)
+-  Collections without a tileset in the tile provider (part 1) [#1011](https://github.com/interactive-instruments/ldproxy/issues/1011)
+-  JSON-FG not updated for the Search module [#1009](https://github.com/interactive-instruments/ldproxy/issues/1009)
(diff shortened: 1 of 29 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-994.surge.sh

cportele:
@azahnen - This should now be ready for review. There is still one outstanding task, but that is a detail and it would be good to check/discuss, if the approach in general makes sense.

azahnen (approved):

PR #1008, opened 2023-07-07
Mapping operations

### LINKED ISSUE #971: Feature Schema: Support representations that better match the application schema

If feature data in an SQL feature provider is based on an application schema that uses inheritance, the current capabilities often cannot properly represent the data in the concepts of the application schema. As a result, the data cannot conform to, for example, a JSON Schema or an XML Schema generated based on an encoding rule.

### Describe the Use Case

Two examples:

1. An application schema has a feature type `Foo` with an attribute `bar` that can be a string, integer or boolean value. In the DB and the feature schema, this could look like this:

```yaml
foo:
  type: OBJECT
  properties:
    bar_stringValue:
      type: STRING
      sourcePath: bar_stringValue
    bar_integerValue:
      type: INTEGER
      sourcePath: bar_integerValue
    bar_booleanValue:
      type: BOOLEAN
      sourcePath: bar_booleanValue
```

However, in a JSON Schema, `bar` will likely be specified as `"bar" : { "type" : [ "string", "integer", "boolean" ] }` and this is not a representation that ldproxy can currently create.

2. An application schema has a feature type `Foo` with an association role `bar` to an abstract class `Baz`, which has multiple subclasses (`Baz1`, ..., `Bazn`). In the DB and the feature schema, this could look like this (using #970):

```yaml
foo:
  type: OBJECT
  properties:
    id:
      type: STRING
      sourcePath: id
      role: ID
    bar_baz1:
      type: FEATURE_REF_ARRAY
      sourcePath: '[id=foo_fk]baz1/id'
    bar_baz2:
      type: FEATURE_REF_ARRAY
      sourcePath: '[id=foo_fk]baz2/id'
    bar_bazn:
      type: FEATURE_REF_ARRAY
      sourcePath: '[id=foo_fk]bazn/id'
```

In a JSON Schema of the application schema, `bar` will likely be specified as a single member/property with an array of all values from the `bar_bazx` arrays. Again, this is not a representation that ldproxy can currently create.

## Overview

Supporting the requirement has two aspects: Add support in the Feature Schema itself for such schemas and add support in the processing of the data. In both use cases, have a more general 'outer' schema of the `bar` property and more specific 'inner properties' with their own specialized schema that are processed to determine the value of `bar` using some operation.

## Type compatibility

Constraints on the types of 'inner properties' depending on the type of the 'outer' property are shown in the table below. A new type `VALUE` is introduced, a generalization of `INTEGER`, `FLOAT`, `STRING`, `BOOLEAN`, `DATETIME`, and `DATE`.

| Outer type  | Valid inner types  | Remarks |
|---|---|---|
| `VALUE`  |  `INTEGER`, `FLOAT`, `STRING`, `BOOLEAN`, `DATETIME`, `DATE`  |   |
| `INTEGER`  |  `INTEGER` |   |
| `FLOAT`  |  `FLOAT` |   |
| `STRING`  |  `STRING`  |   |
| `BOOLEAN`  |  `BOOLEAN`  |   |
| `DATETIME`  |  `DATETIME`  |   |
| `DATE`  |  `DATE`  |   |
| `OBEJCT`  |  `OBJECT`  | Different `objectType` with different schemas can be used  |
| `GEOMETRY`  |  `GEOMETRY`  | Different `geometryType` can be used  |
| `FEATURE_REF `  |  `FEATURE_REF `  | Different `refType` can be used  |
| `VALUE_ARRAY`  |  `VALUE_ARRAY`, `INTEGER`, `FLOAT`, `STRING`, `BOOLEAN`, `DATETIME`, `DATE`  |   |
| `OBJECT_ARRAY`  |  `OBJECT_ARRAY`, `OBJECT`  | Different `objectType` with different schemas can be used  |
| `FEATURE_REF_ARRAY `  |  `FEATURE_REF_ARRAY`, `FEATURE_REF `  | Different `refType` can be used  |

## Operations

Initially, we should support two operations: `coalesce` (first use case) and `concat` (second use case).

### `coalesce`

`coleasce` uses the first value that is not `null`.  `coleasce` would not be supported on arrays.

We can then write the feature schema of the first use case as: 

```yaml
foo:
  type: OBJECT
  properties:
    bar:
      type: VALUE
      coalesce:
      - sourcePath: bar_stringValue
        type: STRING
      - sourcePath: bar_integerValue
        type: INTEGER
      - sourcePath: bar_booleanValue
        type: BOOLEAN
```

### `concat`

`concat` concatenates all values - not the arrays are concatenated, if the type of an inner property is an array, but the items in the array(s). `concat` can only be used on array properties.

We can then write the feature schema of the second use case as: 

```yaml
foo:
  type: OBJECT
  properties:
    id:
      type: STRING
      sourcePath: id
      role: ID
    bar:
      type: FEATURE_REF_ARRAY
      concat:
      - sourcePath: '[id=foo_fk]baz1/id'
        refType: baz1
      - sourcePath: '[id=foo_fk]baz2/id'
        refType: baz2
      - sourcePath: '[id=foo_fk]bazn/id'
        refType: bazn
```

### Additional Information

_No response_

azahnen:
Closes #971.

Known issues:
- [ ] coalesce is disabled for feature generation, schema generation should work
- [ ] queryables endpoint may return unwanted items, warnings are logged
- [ ] unrelated unit tests in xtraplatform-spatial fail with MissingMethodException, disabled for now (they work in IntelliJ)
- [ ] documentation is incomplete

### DIFF
--- a/ogcapi-draft/ogcapi-collections-schema/src/main/java/de/ii/ogcapi/collections/schema/domain/SchemaConfiguration.java
+++ b/ogcapi-draft/ogcapi-collections-schema/src/main/java/de/ii/ogcapi/collections/schema/domain/SchemaConfiguration.java
@@ -11,7 +11,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import de.ii.ogcapi.features.core.domain.JsonSchemaDocument;
 import de.ii.ogcapi.foundation.domain.CachingConfiguration;
 import de.ii.ogcapi.foundation.domain.ExtensionConfiguration;
-import java.util.List;
+import java.util.Set;
 import org.immutables.value.Value;
 
 /**
@@ -35,7 +35,7 @@ public interface SchemaConfiguration extends ExtensionConfiguration, CachingConf
    *     2020-12 (`V202012`), 2019-09 (`V201909`) und 07 (`V7`).
    * @default [ "V202012" ]
    */
-  List<JsonSchemaDocument.VERSION> getVersions();
+  Set<JsonSchemaDocument.VERSION> getVersions();
 
   abstract class Builder extends ExtensionConfiguration.Builder {}
 
--- a/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/domain/FeatureFormatExtension.java
+++ b/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/domain/FeatureFormatExtension.java
@@ -151,6 +151,12 @@ public interface FeatureFormatExtension extends FormatExtension {
                     .map(
                         refType ->
                             String.format("{{apiUri}}/collections/%s/items/{{value}}", refType))
+                    .or(
+                        () ->
+                            property.getConcat().stream()
+                                .filter(p -> p.getRefType().isPresent())
+                                .map(p -> "{{apiUri}}/collections/{{type}}/items/{{value}}")
+                                .findFirst())
                     .orElse(null)));
   }
 
--- a/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/domain/SchemaDeriverJsonSchema.java
+++ b/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/domain/SchemaDeriverJsonSchema.java
@@ -104,6 +104,11 @@ public abstract class SchemaDeriverJsonSchema extends SchemaDeriver<JsonSchema> 
                 property instanceof JsonSchemaArray
                     ? ((JsonSchemaArray) property).getItems()
                     : property)
+        .flatMap(
+            property ->
+                property instanceof JsonSchemaOneOf
+                    ? ((JsonSchemaOneOf) property).getOneOf().stream()
+                    : Stream.of(property))
         .filter(property -> property instanceof JsonSchemaRef && property.getName().isPresent())
         .filter(
             ref ->
@@ -419,5 +424,19 @@ public abstract class SchemaDeriverJsonSchema extends SchemaDeriver<JsonSchema> 
   @Override
   protected JsonSchema withArrayWrapper(JsonSchema schema) {
     return new ImmutableJsonSchemaArray.Builder().name(schema.getName()).items(schema).build();
+  }
+
+  @Override
+  protected JsonSchema withOneOfWrapper(
+      Collection<JsonSchema> schema,
+      Optional<String> name,
+      Optional<String> label,
+      Optional<String> description) {
+    return new ImmutableJsonSchemaOneOf.Builder()
+        .oneOf(schema)
+        .name(name)
+        .title(label)
+        .description(description)
+        .build();
   }
 }
--- a/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/domain/SchemaDeriverOpenApi.java
+++ b/ogcapi-stable/ogcapi-features-core/src/main/java/de/ii/ogcapi/features/core/domain/SchemaDeriverOpenApi.java
@@ -24,6 +24,7 @@ import io.swagger.v3.oas.models.media.Schema;
 import io.swagger.v3.oas.models.media.StringSchema;
 import java.math.BigDecimal;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -290,5 +291,20 @@ public abstract class SchemaDeriverOpenApi extends SchemaDeriver<Schema<?>> {
         .items(oapiSchema)
         .name(oapiSchema.getName())
         .description(oapiSchema.getDescription());
+  }
+
+  @Override
+  protected Schema<?> withOneOfWrapper(
+      Collection<Schema<?>> schema,
+      Optional<String> name,
+      Optional<String> label,
+      Optional<String> description) {
+    Schema<?> schema1 = new Schema<>();
+    schema1.oneOf(new ArrayList<>(schema));
+    name.ifPresent(schema1::name);
+    label.ifPresent(schema1::title);
+    description.ifPresent(schema1::description);
+
+    return schema1;
   }
 }
--- a/ogcapi-stable/ogcapi-features-geojson/src/main/java/de/ii/ogcapi/features/geojson/app/GeoJsonWriterProperties.java
+++ b/ogcapi-stable/ogcapi-features-geojson/src/main/java/de/ii/ogcapi/features/geojson/app/GeoJsonWriterProperties.java
@@ -173,7 +173,12 @@ public class GeoJsonWriterProperties implements GeoJsonWriter {
         }
       } else {
         if (schema.isArray()) {
-          writeValue(json, value, schema.getValueType().orElse(Type.STRING));
+          writeValue(
+              json,
+              value,
+              schema
+                  .getValueType()
+                  .orElse(Objects.requireNonNullElse(context.valueType(), Type.STRING)));
         } else {
           json.writeFieldName(schema.getName());
           writeValue(json, value, schema.getType());
--- a/xtraplatform.gradle
+++ b/xtraplatform.gradle
@@ -2,6 +2,6 @@
 dependencies {
     layers group: 'de.interactive_instruments', name: 'xtraplatform-core', version: '5.3.0-SNAPSHOT'
     layers group: 'de.interactive_instruments', name: 'xtraplatform-native', version: "2.2.0-${platform}"
-    layers group: 'de.interactive_instruments', name: 'xtraplatform-spatial', version: '6.3.0-SNAPSHOT'
+    layers group: 'de.interactive_instruments', name: 'xtraplatform-spatial', version: '6.3.0-mapping-ops-SNAPSHOT'
 }
 


azahnen:
@cportele This can be tested apart from the known issues above. Feel free to add any issues you find.

azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1008.surge.sh

azahnen:
@cportele I fixed the two known implementation issues.

cportele:
@azahnen - I have problems to get it working. I used a working XTrasse configuration and made a minor change to the provider (new FEATURE_REF_ARRAY property that uses `concat` with two source paths). The first source path matches 9 values, the second one 27 values. When I monitor the flow in the feature encoder, I see that for the first 8 values each gets the sequence ARRAY_START, VALUE, ARRAY_END. Then there is ARRAY_START, 28 VALUEs for the remaining 28 values and then ARRAY_END. Only the last 28 values are in the GeoJSON response.

I have updated the `xleit` configuration. The new property is [here](https://gitea.ci.interactive-instruments.de/ldproxy/xleit/src/branch/main/store/entities/providers/xtrassePFS.yml#L11350-L11362). Request: http://localhost:7080/rest/services/xtrassePFS/collections/pfs_plan/items?f=json&skipGeometry=true&profile=rel-as-key&properties=hatPFSObjekt.

azahnen:
@cportele Unlike described in #971, the inner type is currently not optional, so the configuration needs to look like this:

```yaml
hatPFSObjekt:
  type: FEATURE_REF_ARRAY
  concat:
  - sourcePath: "[_id=gehoertzupfs_fk]pfs_alternativetrasseabschnitt/_id"
    refType: pfs_alternativetrasseabschnitt
    type: FEATURE_REF_ARRAY
    valueType: INTEGER
  - sourcePath: "[_id=gehoertzupfs_fk]pfs_gasversorgungsleitungabschnitt/_id"
    refType: pfs_gasversorgungsleitungabschnitt
    type: FEATURE_REF_ARRAY
    valueType: INTEGER
```

I also fixed the `valueType` issue. If the `valueType` is always the same, you can also set it on the outer item:

```yaml
hatPFSObjekt:
  type: FEATURE_REF_ARRAY
  valueType: INTEGER
  concat:
  - sourcePath: "[_id=gehoertzupfs_fk]pfs_alternativetrasseabschnitt/_id"
    refType: pfs_alternativetrasseabschnitt
    type: FEATURE_REF_ARRAY
  - sourcePath: "[_id=gehoertzupfs_fk]pfs_gasversorgungsleitungabschnitt/_id"
    refType: pfs_gasversorgungsleitungabschnitt
    type: FEATURE_REF_ARRAY
```

cportele:
Ah ok, thanks. I will try that.

BTW, I also tried setting `valueType: INTEGER` on the outer item, but then I got an error that concat requires an array, but the type is INTEGER. I will try that again, too.

cportele:
The behavior is changed, but concat still does not work (value type integer now works). 

```yaml
      hatPFSObjekt:
        type: FEATURE_REF_ARRAY
        concat:
        - sourcePath: "[_id=gehoertzupfs_fk]pfs_alternativetrasseabschnitt/_id"
          refType: pfs_alternativetrasseabschnitt
          type: FEATURE_REF_ARRAY
          valueType: INTEGER
        - sourcePath: "[_id=gehoertzupfs_fk]pfs_gasversorgungsleitungabschnitt/_id"
          refType: pfs_gasversorgungsleitungabschnitt      
          valueType: INTEGER
```

Now the values of the first source path come within a single ArrayStart - ArrayEnd block, but for the second source path each value is wrapped in it own ArrayStart - ArrayEnd block. The result is: `hatPFSObjekt": [27]`.

Adding `valueType: INTEGER` on the outer item still results in the following exception: `java.lang.IllegalStateException: Concat may only be used with array types. Found: INTEGER. Path: hatPFSObjekt.`

azahnen:
@cportele Did you pull the recent changes? The first example I posted definitely works for me.

cportele:
I checked and all commits were there. I will check again.

azahnen:
@cportele I double checked that all my changes were pushed and built. I also pushed the configuration changes that work for me.

cportele:
@azahnen - I now also get then correct response for the request - not sure, what part of my setup still had some older state. Anyhow, I have a new problem: I have updated the configuration to add the remaining hatPFSObjekt source paths and also converted hatObjekt.

The `{{type}}` parameter for the collectionId is not replaced based on the source collection, but (it seems) based on the index in the array of the source paths. See http://localhost:7080/rest/services/xtrassePFS/collections/pfs_plan/items?f=html&skipGeometry=true&properties=hatPFSObjekt&profile=rel-as-uri.

cportele:
@azahnen - I have now refactored `pfs_plan` based on what is created by ShapeChange from the model and found the following additional issues:

- `gesetzlicheGrundlage`, `externeReferenz`, `hatObjekt`, `hatPFSObjekt` are no longer included in the response. I guess that is a side-effect from referencing a fragment and then merging additional information for these properties.
- In the configuration created by ShapeChange, the source path for `beteiligte` and `auslegungGemeinden` was set in the type `pfs_plan`, but this results in exceptions (according to the query, the properties from xp_akteur are expected in the pfs_plan table). I had to manually set the source path in the fragment.
- The content of `beteiligte` and `auslegungGemeinden` is an object, not an array as specified in the schema.
- The property name of a property that references a schema in fragments is using the fragment key (e.g., `xp_akteur`), not the property name (`beteiligte`).

azahnen:
@cportele The `{{type}}` issue and fragment key as name issue should be fixed. Looking at the other points now.

azahnen:
@cportele 

> * `gesetzlicheGrundlage`, `externeReferenz`, `hatObjekt`, `hatPFSObjekt` are no longer included in the response. I guess that is a side-effect from referencing a fragment and then merging additional information for these properties.
> * In the configuration created by ShapeChange, the source path for `beteiligte` and `auslegungGemeinden` was set in the type `pfs_plan`, but this results in exceptions (according to the query, the properties from xp_akteur are expected in the pfs_plan table). I had to manually set the source path in the fragment.
> * The content of `beteiligte` and `auslegungGemeinden` is an object (one of the values), not an array as specified in the schema.

I pushed a change to support more complex fragment and merge combinations which should fix the first issue. The second issue sounds like it may be related, so please try that again as well.

cportele:
> I pushed a change to support more complex fragment and merge combinations which should fix the first issue. The second issue sounds like it may be related, so please try that again as well.

The first and fourth issues are resolved as is the `{{type}}` issue. The other two remain:

-  I have to add the source path for `beteiligte` already in the fragment and not only in the partial merge in the type).
- The value is an object and not an array of objects.

azahnen:
> * I have to add the source path for `beteiligte` already in the fragment and not only in the partial merge in the type (otherwise the property is not included in the result).
> * The value is an object and not an array of objects.

They seem to be resolved for me. I pushed the configuration that is working for me, if that is not what ShapeChange generated, please adjust.

azahnen:
@cportele 
> The value seems to be collected correctly, but the result has `href` untransformed (`gehoertZuPlan": { "title": "1", "href": "1" }`) in the `rel-as-link` profile. There is not stringFormat-Tranformer created for the property.

I cannot reproduce that, looks like this for http://localhost:7080/rest/services/xtrassePFS/collections/pfs_alternativetrasseabschnitt/items?f=json&profile=rel-as-link:
```json
"gehoertZuPFS": {
  "title": "Gasanbindung Wilhelmshaven - Leer (GWL)",
  "href": "http://localhost:7080/rest/services/xtrassePFS/collections/pfs_plan/items/1"
}
```

azahnen:
@cportele I pushed a fix.

cportele:
Thanks, @azahnen.

* `beteiligte` and `auslegungGemeinden` now works for me, too.
* `coalesce` and `href`: Works now for me, too.
* For `gesetzlicheGrundlage` and `externeReferenz` I also get an object, not an array. (They were missing, because I had dropped the `schema` in the "secondary" listing in the type with the `sourcePath`. I have added them again and pushed the change. They are now included again, but with the incorrect content.)

If I haven't lost the overview, the last issues is currently the only remaining issue that I am aware of (beside the open issues in the first comment).

cportele:
I think we need to increase the levels of `schema` references. I just looked at a type that has a chain of four `schema` references until the root supertype. The properties of that last root type fragment are not included in the response, so I assume they are not processed.

azahnen:
@cportele There is a fundamental issue with the generated schema, multi-level merge is currently not supported. For example given this

```yaml
  bst_erdgasleitung:
    type: OBJECT
    objectType: BST_Erdgasleitung
    label: BST_Erdgasleitung
    description: Erdgasleitung im Bestand
    merge:
    - schema: '#/fragments/bst_multi_linienobjekt'
    - properties:
        druckstufe:
          sourcePath: druckstufe_fkcl
          type: STRING
          label: druckstufe
          description: Angabe der Druckstufe.
          constraints:
            codelist: XP_DruckstufeGas
```

the merge in `bst_multi_linienobjekt` will be ignored upon resolution, since the items in merge are of type `PartialObjectSchema`, which does not have the merge option. 
I have to think about that.

azahnen:
To make it clearer, after the resolution the example would look like this:

```yaml
bst_erdgasleitung:
    type: OBJECT
    objectType: BST_Erdgasleitung
    label: BST_Erdgasleitung
    description: Erdgasleitung im Bestand
    merge:
    - merge:
      - schema: '#/fragments/bst_objekt'
      - properties:
        position:
          sourcePath: position
          type: GEOMETRY
          geometryType: MULTI_LINE_STRING
    - properties:
        druckstufe:
          sourcePath: druckstufe_fkcl
          type: STRING
          label: druckstufe
          description: Angabe der Druckstufe.
          constraints:
            codelist: XP_DruckstufeGas
```

Since merge items don't have merge, it is silently ignored.

cportele:
Ah, that explains the behavior.

With `allOf` we had used nested `allOf`s; for example, [here](https://github.com/ldproxy/demo/blob/main/store/entities/features/withTiles/providers/cologne_lod2.yml#L25-L54). Does that mean that this won't work now.

azahnen:
@cportele Nested inside of properties is fine, the issue is directly nested. 

I pushed another fix and adjusted the configuration to not use merge, since I don't think any of them is needed really. That way it at least seems to work as expected.

cportele:
@azahnen - Great! I think we do have an XTrasse configuration that works as expected and we can now adjust the ShapeChange target.

For cases where one needs to merge a schema and some additional properties one really never needs the `merge`, the implicit patch is always sufficient? We really only need `merge` when there are multiple schemas that should be merged or if locally a property from a schema needs to be "overloaded" / replaced?

azahnen:
Or if one of the items has a different parent source path, for example the json column case. But that is more relevant for external schemas or when not using schema at all, with fragment s you can also define it there.

azahnen:
@cportele I fixed the issue with the outer valueType and also removed the need for the inner type. I simplified the configuration in main accordingly.

cportele:
I updated also the sc branch. It works as expected. 👍 

I will try tests in some other configurations to cover a few additional cases, but I think we are close to merging this.

azahnen:
@cportele That is not supported, neither concat nor coalesce support mixing values and feature refs. See the type compatibility table in #971. I added some checks for unsupported cases.

The other issue is that the last sourcePath is missing a column, which obviously was not catched by any check but led to a wrong query. 

This works as expected:

```yaml
      result: 
        type: VALUE
        label: Ergebnis
        coalesce:
        - sourcePath: result_measurementsingleobs
          type: FLOAT
        - sourcePath: result_measurementpairobs_1
          type: FLOAT
        - sourcePath: result_measurementpairobs_2
          type: FLOAT
        - sourcePath: result_measurementtripleobs_1
          type: FLOAT
        - sourcePath: result_measurementtripleobs_2
          type: FLOAT
        - sourcePath: result_measurementtripleobs_3
          type: FLOAT
        - sourcePath: result_textobs
          type: STRING
        - sourcePath: result_matrixobs
          type: STRING
        - sourcePath: '[id=observationid]observation_result_classificationobs{sortKey=seqno}/[classificationcodeid=code]classificationcode/code'
          type: VALUE_ARRAY
          valueType: STRING
```

Note that transformations for concat/coalesce items are currently not supported (only for nested properties in objects). So you couldn't even create an uri with stringFormat in this case.
You could create a wrapper object like this:

```yaml
      result: 
        type: OBJECT
        label: Ergebnis
        coalesce:
        - type: OBJECT
          properties:
            value:
              sourcePath: result_measurementsingleobs
              type: FLOAT
        - type: OBJECT
          sourcePath: '[id=observationid]observation_result_classificationobs{sortKey=seqno}/[classificationcodeid=code]classificationcode'
          properties:
            title:
              sourcePath: symbol
            href:
              sourcePath: code
              transformations:
                stringFormat: '{{apiUri}}r/collections/classificationcode/items/{{value}}'
```

cportele:
@azahnen - thanks for the reminder about type compatibility, I had forgotten about this when I created the config.

Also thanks for the ideas, I will use them as a starting point for further tests.

Issue #721, opened 2022-09-10
add auth flow integration


azahnen:
Currently ldproxy just expects a bearer token to be passed in the `Authorization` header. There are no automatic redirects to the login form of the identity provider. 
This will integrate html views and all clients applications provided by ldproxy like swagger ui with the authentication flow of the identity provider.

PR #1031, opened 2023-08-08
Rbac

### LINKED ISSUE #720: add fine-grained role based access control
In #718 basic rbac with roles `read` and `write` was introduced. 
This will add additional roles that allow fine-grained access control down to the operation level.
Role is used in a broader sense here, you could also call them permissions. The idea is that a user can have any combination of roles. The list of roles is part of the token/userinfo and can be evaluated in the authorization layer of ldproxy.

#### scopes

OAuth scopes are used when an end-user wants to allow a third-party application to access an API on his behalf. The user is then presented a consent dialogue with the permissions the application requests, which he has to confirm and may be able to change. Therefore the scopes should be broad categories of permissions that are understandable to an end-user.

This will add support for OAuth scopes, but these broad categories of permissions will also be used as the first group of new roles. A reasonable use-case oriented categorization already exists, the tags used in the OpenAPI definition. With some slight changes, the list of scopes/roles would be:

- `explore` or `metadata:read`? *Capabilities*, *Discover data collections*, maybe the discovery parts of the other scopes
- `data:read` *Access data*
- `data:write` *Mutate data*
- `styles:read` *Discover and fetch styles*
- `styles:write` *Create, update and delete styles*
- `tiles:read` *Discover and fetch tiling schemes*, *Access multi-layer map tiles*, *Access single-layer map tiles*
- `resources:read` *Discover and fetch other resources*
- `routing:read` *Routing*

#### operation roles

In addition to the scopes, it should be possible to restrict access on the OpenAPI operation level. An operation is the  combination of an endpoint and a method. Every operation should have an id, which will be added if not present, that is used as role name.
All operation roles belong to a scope. The roles are organized in a hierarchy like this:

- `read`
  - `data:read`
    - `data:viewItem` *GET* */{api}/collections/{collection}/items/{id}*
    - ... 
  - `styles:read`
  - ...
- `write`
  - `data:write`
    - `data:replaceItem` *PUT* */{api}/collections/{collection}/items/{id}*
    - `data:deleteItem` *DELETE* */{api}/collections/{collection}/items/{id}*
    - ...
  - `styles:write'
    - ... 
  
A user with role `write` would have access to all writing operation on data and styles.
A user with role `data:write` would only have access to writing operation on data.
A user with role `data:replaceItem` would only have access to a single operation, so replacing an item in this case.

#### entity roles

All of the roles described above are static, they are the same for each deployment. They define what a user is allowed to do, but not on which entities he is allowed to do it. The role `data:viewItem` would allow the user to view items in any collection of any API.
This adds dynamic roles for API and collection ids. A user needs a combination of at least one operation role and at least one entity role to be allowed to do anything. That might be achieved with the help of the configurable public roles, see below. 

- `allApis` access any API and collection
- `daraa` access any collection in API *daraa* 
- `daraa:AeronauticSrf` acess collection *AeronauticSrf* in API *daraa*

Since an Identity Provider might be responsible for multiple deployments of ldproxy and multiple deployments might have APIs with the same id, it might be necessary to add an option to define a deployment id that is used as prefix for the entity roles.

#### configuration

This will slightly change the default configuration defined in #718, so that by default anyone is allowed any reading operation on any API:

```yaml
accessControl:
  mode: RESTRICTED
  public:
  - read
  - allApis
```

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #720.

This can be tested by defining permissions for the role `public`, see the updated docs for details. 

Please review if the assignment of operations to scopes makes sense.

As part of #721, the potential scopes/permissions to access an operation will be added to the OpenAPI definition.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1031.surge.sh

cportele:
@azahnen - Initial observations after first tests:

- Accessing operations not in the public scope (with or without API/collection qualifier) list returns 401 as expected.
- The OpenAPI definition is not yet updated and does not reflect the access control configuration.
- TODO: Review the list of scopes and their associated operations.

cportele:
@azahnen 

Comments on the mapping of operations to scopes as well as the naming of scopes:

* `GET` `/conformance` should be in `discover`
* `metadata` is not intuitive, there is a lot of metadata in other scopes, too; proposal: `collections`
  * Since there will be future operations to create new collections, I would directly use `collections:read`  
* `discovery` is needed for a generic client to make use of the API. All operations in `metadata`/`collections` are de-facto prerequisites for `data:read` (and some operations are needed for `data:write`); for discussion: Should, e.g., `data:read` include `discover` etc.?
* `GET` `/resources` and  `/resources/{file}` -> `resources:read` and `resources`
* `PUT` `/resources/{file}` -> `resources:write` and `resources`
* `POST` `/routes` is not `routes:read` as it is not related to stored routes; maybe something like `routes:execute`?
* `POST` `/search` should be in `data:read` (unrelated to stored queries)
* `GET` `/search/{queryId}` really belongs to `data:read` as it returns features, not information about stored queries, but it really only is usable, if `queries:read` is enabled; maybe `queries:execute`? In that case, `POST` `/search` could also be in `queries:execute`
* Since the module is called SEARCH, maybe change `queries:*` to `search:*`?

It is probably best to discuss in a call.

azahnen:
@cportele I implemented most of the proposed changes.

> Should, e.g., `data:read` include `discover` etc.?

That is currently not possible, inclusion is based on common tokens. So `data` really means `data:*` and `read` really means `*:read`.

cportele:
@azahnen - OK, we should then just document that if a role includes `data:read`, then the role should also include `discover` and `collections`. Actually, almost any role will have to include `discover` unless that role is expected to submit specific requests based on out-of-band information.

azahnen:
@cportele I would expect `discover` to be in `public` in most cases. Then it does not have to be repeated in any other roles.

cportele (approved):

PR #1042, opened 2023-08-18
deprecate postOnItems

### LINKED ISSUE #1004: Deprecate postOnItems?
Some time ago, we created the option to submit queries with POST on `/collections/{collectionId}/items` to support large filter expressions (e.g., large geometries) that result in URIs that are too long. Module `FEATURES_EXTENSIONS`, option `postOnItems`.

With the introduction of the `SEARCH` module this should no longer be necessary, as that module provides the option for ad-hoc queries via POST in a standardized manner. 

POST on `/collections/{collectionId}/items` will not be standardized for queries, only to create new features. While it is possible to support both at the same time in an API, with the behavior of the operation depending on the content type, it is considered too confusing.

Should we mark `postOnItems` as deprecated with v3.5, suggest to use `SEARCH` instead and remove the option in v4.0?

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1004.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1042.surge.sh

azahnen (approved):

PR #1024, opened 2023-08-02
add support for terrain in MapLibre styles

cportele:
### Pull request checklist

-   [ ] Added/updated unit tests
-   [ ] Updated documentation
-   [x] All checks are passing

### Changes introduced by this PR

Closes #1023.

PR #1044, opened 2023-08-18
fix issues in the derived style for feature(s)

cportele:
### Pull request checklist

-   [ ] Updated documentation

This is a follow-up to PR #1024 that added support for terrain in MapLibre styles. It turned out that for some reason, using a terrain in a default style led to JavaScript errors when displaying features with the derived style. This change removes terrains from the derived styles for the GeoJSON features.

In addition, hillshade and background layers were not properly handled when a style for the GeoJSON features were derived for displaying feature(s) in the HTML representation.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1044.surge.sh

azahnen (approved):

PR #1045, opened 2023-08-18
CRUD improvements (continued)

cportele:
### Pull request checklist

-   [x] Updated documentation

This is a follow-up to PR #994, which turned out to be incomplete with respect to the requirements in the current draft of OGC API Features Part 4:

- The conformance class URI(s) for the optimistic locking conformance classes had not been updated.
- If optimistic locking is enabled, requests without the relevant header (If-Match, If-Unmodified-Since) must be rejected.
- So far, both conformance classes were always enabled. However, this will not be desired in all cases; for example, if concurrent editing is not an issue or if no last-modified information is available. The conformance classes must now be enabled explicitly.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1045.surge.sh

azahnen (approved):
I did not check the draft, but wouldn't it be more fitting to throw `400` in `checkHeader`?

cportele:
@azahnen  - `IllegalArgumentException` is mapped to a `400`, but in an endpoint class we can also throw the web exception directly. I will change the exception.

PR #1047, opened 2023-08-31
Security

### LINKED ISSUE #721: add auth flow integration

### LINKED ISSUE #722: add attribute based access control
A very basic support for XACML PDPs was introduced in #718. It mainly covers use cases that can be solved much easier with the fine-grained rbac that will be introduced in #720.
The only real abac case it covers is creating or replacing features, where the body is passed as string in the XACML request. Even that is not that useful, since you cannot use e.g. GeoJSON properties or geometries directly in conditions. You only get the body as string, so you are mostly limited to defining conditions with `string-contains`.

This will introduce real abac that allows you to define conditions based on user attributes, feature attributes and feature geometries (if the PDP supports GeoXACML, only known OSS Implementation is AuthZForce). 
In the first place this means that the PDP needs a way to access these informations which are currently not part of the XACML request. There are two approaches:

- **pull** 
  This is essentially the PIP (Policy Information Point) pattern. The idea is that you can just request the needed attributes in your policies and the PDP knows how to get them by sending requests to ldproxy. This approach has two drawbacks. 
  First, this requires writing custom extensions for the PDP, which is possible and relatively easy for the reviewed OSS PDPs (WSO2 + AuthZForce), but it binds you to the chosen PDP and it is a separate library that has to be maintained. 
  Second, since in this scenario access to ldproxy is restricted, the PDP would itself need an authorization to access the needed data. So most likely there would a be user with unrestricted read access whose credentials would have to be put into a configuration file of the PDP.
- **push**
  The idea here is that ldproxy already knows which attributes might be used in the policy and puts them into the XACML request. That means the API configuration will need a new option to define e.g. which properties of a feature will be send to the PDP when the user wants to view a feature by id.
  This is the preferred approach, since the new code would be maintained as part of ldproxy, there is no additional communication needed between PDP and ldproxy and it works with any PDP.

Additionally this will add support for XACML obligations that allow to send back additional or adjusted query parameters back to ldproxy in the XACML response. This is useful e.g. if a user wants to view all features in a collection but is only allowed to access a subset. Instead of denying the access completely it could be permitted with the obligation that the API response has to be filtered, so that the user will only see the allowed features.

### LINKED ISSUE #925: add audience claim support
This will allow to configure a specific audience for each ldproxy deployment. If an audience is configured, only access tokens with an [audience claim](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3) that contains the given audience are accepted.

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #721.
Closes #925.
Closes #722.

There are also some changes to OpenAPI and the swagger-ui integration:

- removes `ApiDefinitionAuxiliaryFormatExtension` and `EndpointDefinitionFile`, uses a mustache-based integration instead
- ~~updates OpenAPI media types, see https://datatracker.ietf.org/doc/draft-ietf-httpapi-rest-api-mediatypes/~~ reverted
- adds common ldproxy header and footer to API documentation

Merge after `migration`.
Depends on interactive-instruments/xtraplatform#135.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1047.surge.sh

PR #1053, opened 2023-09-15
Support queryables in nested arrays

### LINKED ISSUE #1049: Support queryables in nested arrays
The resolution of issue #431 was incomplete. It only mapped VALUE_ARRAY properties to a array queryable. If the queryable property was a simple value, but nested in an OBEJCT_ARRAY, the queryable is still a VALUE (STRING, etc.). These queryables, however, should also be a VALUE_ARRAY. 

To determine the type of a queryable, the parent properties also need to be investigated.

Note that this does not affect sortables since sortables cannot be arrays or in arrays.

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1049.

Depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/223.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1053.surge.sh

PR #1057, opened 2023-09-21
add specification information to API components

### LINKED ISSUE #1056: update OpenAPI definition to state maturity of API components
The maturity of each building block is currently documented in the documentation. However, the documentation is accessed by API authors. Developers of API clients / API users will typically access the OpenAPI description to learn about the API. 

The maturity of each API component (in particular: operation and query parameter, but also path parameter, header) should be included in the description - and perhaps also in an extension element (e.g., `x-maturity`).

Since this will typically affect a large number of API components in an OpenAPI definition, the additional description should be brief, consistent across all components and easy to spot for users in Swagger UI.

Approach:

- Distinguish five maturity values:
  - `STABLE_OGC`: from an approved OGC Standard
  - `DRAFT_OGC`: from a draft OGC Standard, proposal or similar
  - `STABLE_LDPROXY`: ldproxy-specific extension in ogcapi-stable (or ogcapi-custom, if considered stable)
  - `DRAFT_LDPROXY`: ldproxy-specific extension in ogcapi-draft (or ogcapi-custom, if not considered stable)
  - `DEPRECATED`: could have formerly been any of the above, but is now marked as deprecated
- Add two new properties to `ApiExtension` and `ApiOperation`:
  - `specificationMaturity`: see above for the values
  - `specificationRef`: URL and Titel of the specification (for OGC documents)
- Set these values for all operations, path parameters, query parameters, and headers
  - Note: It would be good to add these also for the formats, but I see no way to express the maturity of any media type in the OpenAPI definitions, so I have not set these for now.
- In `FOUNDATION` a new option `includeSpecificationInformation` has been added to control, if such information is added to the API definition. Default is `false`.
  - Note: The default is `false` for two reasons. First, be consistent with current behavior. Second, for APIs that only use stable building blocks, such information will typically be confusing.
- The following is added in the OpenAPI definition, if the option is enabled:
  - The following paragraph is added to the API description: _Note: This is API is based on API building blocks (e.g., operations, query parameters, or headers) specified in OGC API Standards or drafts of those standards. For more information about OGC API Standards, see https://ogcapi.ogc.org/. Some building blocks of this API can be preliminary and may change in future versions of this API, because they are not yet based on a stable specification. The maturity is stated for each building block._
  - A line is added at the end of the description of the operation/parameter/header depending on the maturity value:
    - Maturity: `STABLE`
    - Maturity: `PRELIMINARY` (specified in a draft standard; the parameter may change in future versions of this API)
    - Maturity: `PRELIMINARY` (the header may change in future versions of this API)
    - Maturity: `DEPRECATED` (the operation may be removed in future versions of this API)
  - Also an extension element `x-maturity` is added.
  - For deprecated components, this information is also set in the OpenAPI definition.
  - If no other external document has been configured for the component,  the specification reference is added as `extenalDocs`.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1056.

See the first comment in #1056 for a description of the changes.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1057.surge.sh

azahnen (approved):

PR #1060, opened 2023-09-27
Custom html

### LINKED ISSUE #1003: improve html customization
Currently overriding the mustache templates is often not a sustainable approach, since the templates contain a lot of things that are irrelevant for most customizations, but which will cause issues when the original templates change on an update.

This would identify useful extensions points to add e.g. logos, additional links, stylesheets, javascript etc. and add empty templates that are included at these extension points. That will allow to override these empty templates wihtout any issues on updates.

This would also add the possibility to publish global assets like images and stylesheets. Currently that is only possible as part of an API.

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #1003.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1060.surge.sh

cportele (approved):
Looks good

PR #1081, opened 2023-10-18
Tiles: deprecate datasetTiles, collectionTiles

### LINKED ISSUE #1005: Collections without a tileset in the tile provider (part 2)
An issue exists in the configuration of tiles.

Whether tiles are enabled for a collection is currently determined by the option `enabled` of the `TilesConfiguration` of the collection. In addition, there is an option `collectionTiles` that can be set to `false` to disable tiles for all collections without the need to explicitly disable tiles for every collection. 

With the latest redesign of the tiles module, the tile providers were introduced and for tile providers from feature data, the collections for which tiles are created is now also configured in the tile provider.

As a result, currently both the tilesets in the tile provider and the collection in the services configuration have to be changed if a collection / feature type should be added or removed from the tiles. This is prone to errors and also very tedious in APIs with dozens of collections.

Probably the best way to improve this is to deprecate the `datasetTiles` and `collectionTiles` option and to consider TILES enabled for a collection only, if there is a tileset in the tile provider.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1005.

If a tile provider exists, the provider is used to determine, which tilesets exist. Otherwise the deprecated options are used (until v4).


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1081.surge.sh

PR #1085, opened 2023-10-21
support properties that are neither returnable nor receivable

### LINKED ISSUE #1084: support queryables/sortables that are not returnables
There are situations where

- some property should not be included in the feature response, e.g., because it is not part of a schema specified by a community,
- but there is a need/wish to filter on the property.

Example: Accident features reference the road segment feature on which they are located. The road class is a property of the road segment, but it is not included in the representation of the accident features. However, if there is a need to filter accidents by road class, a queryable "roadClass" has to be specified, but "roadClass" cannot be a returnable property.

Such properties have to be defined in the feature schema just like any other property, but they have to be annotated as properties that are not returnables. Note that there may be cases where such properties have scope MUTATIONS.

In order to be usable as a queryable or sortable, they have to be included in the table schema (SchemaSql) when generating the query. However, the properties that are not returnable have to be removed from the value queries and schema associated with the SQL row. 

These properties also have to be removed from the schema resource of the collection unless they have scope MUTATIONS.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1084.

Depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/231.

One way to test this is to change the `unfaelle` type in the `strassen` API:

- Annotate `abs` with `excludedScopes: [RETURNABLE, RECEIVABLE]`. (This, of course, does not really make sense, just for testing.)
- Add a new property:
  ```yaml
      strassenklasse:
        sourcePath: "[abs=abs]abschnitteaeste_line/strkl"
        type: STRING
        label: Straßenklasse
        constraints:
          enum:
          - A
          - B
          - L
          - K
        excludedScopes:
          - RETURNABLE
          - RECEIVABLE
  ``` 
- Add `abs` and `strassenklasse` to the list of queryables.
- Add `abs` to the list of sortables.

Then:

- http://localhost:7080/rest/services/strassen/collections/unfaelle/schema excludes both `abs` and `strassenklasse`.
- http://localhost:7080/rest/services/strassen/collections/unfaelle/items does not include `abs` and `strassenklasse`.
- http://localhost:7080/rest/services/strassen/collections/unfaelle/items?abs=5209062A5209047O selects all accidents on the road segment.
- http://localhost:7080/rest/services/strassen/collections/unfaelle/items?strassenklasse=A&sortby=-abs select all accidents on motorways, sorted by the segment id in descending order.


azahnen (changes requested):
Like already discussed, in general this looks good, but before merging we should harmonize the usage of flags and scopes in the `FeatureSchema`.

cportele:
I still want to do more testing, so I marked this as a draft for now.

cportele:
I have updated the PRs. `getScope()`, `getIsQueryable()`, and `getIsSortable()` are now deprecated. 

Instead, `getExcludedScopes()` has been added to exclude a property from the Returnables, Receivables, Queryables, or Sortables. This approach is used over an option `getScopes()` to avoid issues, for example, with incomplete schemas that can occur when `merge` is used in a feature schema.

cportele:
@azahnen - I have added the check method and adjusted the remaining code. 

I only added the added the check method in FeatureSchema, not the other two immutable subtypes of SchemaBase: FeatureSchemaExt and SchemaSql:

- SchemaSql is always built from a FeatureSchema as far as I can tell, so this should not be necessary. 
- FeatureSchemaExt does not seem to be used? If that is the case, we should delete the class.

PR #1087, opened 2023-10-23
s3 and value store

### LINKED ISSUE #878: support s3 store sources
This will add support for S3 compatible store sources.

### LINKED ISSUE #1075: DEBUG log messages despite logging.level set to INFO
### ldproxy Version

3.5.0.

### Current Behavior

In cfg.yml logging.level is set to INFO. However, in xtraplatform.log DEBUG messages are logged. 
Find cfg.yml attached and an excerpt from xtraplatform.log below.

[cfg.txt](https://github.com/interactive-instruments/ldproxy/files/12871259/cfg.txt)

### Expected Behavior

Log-entries to xtraplatform.log respect logging.level.

### Steps to Reproduce

- start ldproxy with cfg.yml
- navigate through an API 
- check logs for DEBUG messages

### Relevant log output

```shell
INFO  [2023-10-11 16:37:32,634] zulaessige_hoechstgeschw - Service with id 'zulaessige_hoechstgeschwindigkeiten' started successfully.  
DEBUG [2023-10-11 16:42:10,995]                          - Processing request: GET /rest/services [a912e9cf-f3c0-40c0-8988-c66f2fd281d4] 
DEBUG [2023-10-11 16:42:11,052]                          - Sending response: 200 OK [a912e9cf-f3c0-40c0-8988-c66f2fd281d4] 
DEBUG [2023-10-11 16:42:14,334] stadtgruen               - Processing request: GET / [c5e23f1a-a28e-4757-b09c-aa4c7ce18670] 
DEBUG [2023-10-11 16:42:14,373] stadtgruen               - Sending response: 200 OK [c5e23f1a-a28e-4757-b09c-aa4c7ce18670] 
DEBUG [2023-10-11 16:42:17,433] stadtgruen               - Processing request: GET /collections [9439c082-eeea-4daa-9491-84bbc4d42116] 
DEBUG [2023-10-11 16:42:17,475] stadtgruen               - Sending response: 200 OK [9439c082-eeea-4daa-9491-84bbc4d42116]
```

### LINKED ISSUE #1077: improve storage strategies
This is a re-evaluation of the storage strategies/categories that the store implementation provides to consumers, taking into account remote store sources and multi-node setups.
  
#### Criteria
- have a unique id
- may have defaults and overrides
- are only read on startup or explicit reload
- are never written by the application
- are used as objects in the application
- represent main components of the application

#### Types
- providers
- services

#### Caching
- the objects are cached in memory
- the source files are never cached

#### Assessment
- no changes needed

#### Criteria
- are identified by their path
- are passed through to clients on demand
- may be written by the application

#### Types
- tiles (plain)
- tiles3d
- api-resources
- json-ld-contexts
- html/assets

#### Caching
- never cached by the application
- should be cached by clients using etags etc.

#### Assessment
- no changes needed
- write operations should work as-is in multi-node setups
- tile seeding in multi-node setups will need a synchronization mechanism, e.g. based on redis, but that is to synchronize the work, not the storage

#### Criteria
- are identified by their path
- are provided to third party libs as local files on demand
- may be written by the application, but only for FS sources

#### Types
- features (gpkg)
- tiles (mbtiles)
- proj

#### Caching
- for sources other than FS, these are cached by the store
- if cached, changes are only detected on application restart

#### Assessment
- no changes needed

## TBD1
These are currently entities or blobs but do not match the criteria exactly. They should get a new storage type.

#### Criteria
- have a unique id
- are only read on startup or explicit reload
- are never written by the application
- are used as objects in the application

#### Types
- codelists
- tile-matrix-sets

#### Caching
- the objects are cached in memory
- the source files are never cached

#### Assessment
- These could go to entities, but be marked as auxiliary, which means no defaults and overrides and no need for separate data and entity classes in the code.
- Or they could go to TBD2, which might be the better option.

## TBD2
These are currently blobs, but that does not really fit. They should get a new storage type.

#### Criteria
- are identified by their path
- are read on demand
- may be written
- are used as objects in the application
- are passed through to clients

#### Types
- queries
- routes 
- styles (and metadata)

#### Caching
- currently none, but needed for remote sources
- either file or in-memory, could also be decided per type in the implementation
- changes are only detected on application restart or explicit reload

#### Assessment
- the new storage type might be called `values`
- it would use the same (de)serialization as entities, so YAML and JSON are supported, but no defaults or overrides and an id is not required 
- storage paths would consist of a type prefix, an arbitrary path and name and the format suffix, e.g. `values/styles/daraa/night.json`
- the store encoding might differ from the encoding passed to clients, e.g. in the case of styles, we could just store MapLibre styles, or we could come up with a simplified or specification-agnostic syntax from which the MapLibre styles and potentially other style formats could be derived
- if we plan to do the latter in the future, we should include the format/spec in the type, e.g. `maplibre-styles`, so we could introduce `styles` for a generic format later
- style metadata would need a separate type, e.g. `style-metadata` (could also be moved to blobs since it is only passed through as far as I see, but splitting styles and metadata might lead to issues)
- manual changes to stored files by users would no longer be detected automatically, so we would need a (re)load task like we have for entities
- write operations in a multi-node setup would need a synchronization mechanism, e.g. based on redis, which should be addressed along with [\#814](https://github.com/interactive-instruments/ldproxy/issues/814)

## TBD3
These are currently blobs but do not match the criteria exactly. 

#### Types
- html/templates
- schemas

#### Assessment
- html/templates are passed to a third party lib, but are not strictly needed as local files. But since they may be requested often and are never written, they should be moved to local blobs.
- schemas are passed to a third party lib, but they are read only once on provider startup and are not needed as local files. They should stay in blobs.

### LINKED ISSUE #1086: upgrade PROJ from 9.1.0 to 9.3.0

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #878.
Closes #1077.
Closes #1086.
Fixes #1075.

Depends on interactive-instruments/xtraplatform#144.
Depends on interactive-instruments/xtraplatform-native#3.
Depends on interactive-instruments/xtraplatform-spatial#232.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1087.surge.sh

cportele (approved):
I did some tests including accessing parts of the configuration from GitHub. I think we should merge it into master. I will do more tests in the next days.

There is one aspect that needs to be addressed (along with the documentation): The special style support is only for maplibre styles, but for 3D Tiles and Cesium, the 3D Tiles styles have the same role.

PR #1091, opened 2023-10-31
Transformation pipeline

### LINKED ISSUE #976: Property transformation for profile 'rel-as-link'
This is a follow-up to:

- #972
- #975 

`rel-as-link` is not yet fully supported. HTML and GML are supported using property transformation, but mapping a value to a Link object requires more complex changes to the logic in `FeatureTokenTransformerSchemaMappings`. As a workaround, the mapping to a Link object is for now implemented in the GeoJSON feature encoder. This should be replaced by a proper transformation to create a Link object.

### LINKED ISSUE #977: Support title in feature reference links
This is a follow-up to:

- #972
- #975

It should be possible for the profile `rel-as-link` to include a title from different source path(s), otherwise this option will not be useful for cases where the feature ID has no meaning (e.g., for auto-increment ids).

This will likely require that the feature reference is defined as an object with properties for the id and the title. The configuration described in #970 will continue to be supported.

### LINKED ISSUE #1071: Incorrect handling of object array properties
### ldproxy Version

3.5.0

### Current Behavior

It seems that a problem was introduced in #1008. The behavior is correct in 3.4.2, but not in 3.5.0.

Example: 

API: `demo/cologne_lod2`
Request: `/collections/building/items/DENW37AL1000qrbQ?f=jsonfg`

The building has two building parts. In 3.4.2 the events are:

```
...
START ARRAY consistsOfBuildingPart
START OBJECT consistsOfBuildingPart [1, 1]
...
END OBJECT consistsOfBuildingPart [1, 1, 13]  
START OBJECT consistsOfBuildingPart [1, 2]
...
END OBJECT consistsOfBuildingPart [1, 2, 6]
END ARRAY consistsOfBuildingPart
...
```

In 3.5.0, the events are:

```
...
START ARRAY consistsOfBuildingPart
START OBJECT consistsOfBuildingPart [1, 1]
...
END OBJECT consistsOfBuildingPart [1]
END ARRAY consistsOfBuildingPart
...
```

The second START OBJECT for the second building part object in `FeatureTokenTransformerSchemaMappings` does not result in closing the first building part object and opening the second building part object. The reason is that `context.indexes()` in the `NestingTracker` is no longer `[1, 1, 13]`, but just `[1]`.   

The result is a incorrect and incomplete building feature.

### Expected Behavior

All building parts are present.

### Steps to Reproduce

See above.

### Relevant log output

_No response_

### LINKED ISSUE #1076: coalesce and arrays
### ldproxy Version

3.5.0

### Current Behavior

`coalesce` does not work in an array. The following feature property "foo" always returns an array with "bar" only set in the first array item. When the first "foo.bar" value is processed, the path "foo.bar" is added to `MappingOperationsTransformer.coalesces`. As a result, all subsequent "foo.bar" values are skipped in `MappingOperationsTransformer.valueCoalesce()`.

```yaml
foo:
  type: OBJECT_ARRAY
  sourcePath: "[id=id]foo_table"
  properties:
    baz:
      type: STRING
    bar:
      type: VALUE
      coalesce:
      - sourcePath: bar_stringValue
        type: STRING
      - sourcePath: bar_integerValue
        type: INTEGER
      - sourcePath: bar_booleanValue
        type: BOOLEAN
```

### Expected Behavior

All values are returned, not only the first one.

### Steps to Reproduce

See above.

### Relevant log output

_No response_

### LINKED ISSUE #1089: support re-sorting of json column properties
Property re-sorting for SQL feature providers is mainly established by the query builder, some edge cases are handled in a sorting transformer. 
Properties from [JSON columns](#907) cannot be re-sorted yet, they are passed through in the given order.

This will implement an improved sorting transformer that always enforces the order defined in the feature schema. The transformer could also be used for other feature provider types which do not support property re-sorting yet.

### LINKED ISSUE #1090: revise feature transformation pipeline
Some parts of the feature transformation pipeline have become very hard to test and maintain. The recent addition of [mapping operations](#971) introduced some issues in what should have been unrelated parts of the pipeline. There are also some new requirements that cannot be implemented in a reasonable way using the current design. This revision will address the pain points and incorporate the new requirements.

### Before

Image

### After

Image

### Main changes 

- move the parts of the schema mapping transformer that operate directly on feature schemas outside of the pipeline
- remove the context value buffer, which was mainly (mis)used as a crutch when something did not fit into the pipeline concept
- remove the related context transformers
- introduce a new token buffer as part of the schema mapping transformer that is used for operations that do not really fit into the pipeline concept
- introduce token slice transformers which operate on a slice of the token buffer

### Related issues and enhancements

This revision should fix the following bugs: 

- coalesce and arrays (#1076)
- incorrect handling of object array properties (#1071)

Furthermore it should implement or provide an easy way to implement these enhancements:

- property transformation for profile 'rel-as-link' (#976)
- support title in feature reference links (#977)
- support re-sorting of json column properties (#1089)

azahnen:
### Pull request checklist

-   [ ] Updated documentation

- [x] fixes #1076
- [x] fixes #1071
- [x] closes #976
- [x] closes #977
- [x] closes #1089
- [x] closes #1090

Depends on interactive-instruments/xtraplatform-spatial#233.


azahnen:
@cportele

Concat, coalesce and flattening are currently **not** working. Everything else should work as expected, the improved sorting and feature refs can be tested.

I decided not to introduce `FEATURE_REF_OBJECT` and `FEATURE_REF_OBJECT_ARRAY` since 

- It would add a lot of unecessary complexity. 
- `FEATURE_REF` is actually already an object, but we do not treat it as one, which is why we need special code to transport the `refType`.
- The actual encoding is solely driven by the profile/format, so having two different types for the same thing in the feature schema would be counterintuitive.


That means the previous syntax 

```yml
gehoertZuPlan:
  type: FEATURE_REF
  valueType: INTEGER
  sourcePath: [gehoertzuplan_pfs_plan_fk=_id]pfs_plan/_id
  refType: pfs_plan
```

is still supported, but it is just a shortcut for

```yml
gehoertZuPlan:
  type: FEATURE_REF
  sourcePath: [gehoertzuplan_pfs_plan_fk=_id]pfs_plan
  refType: pfs_plan
  properties:
    id: 
      type: INTEGER
      sourcePath: _id
    title: 
      type: STRING
      sourcePath: _id
```

which is internally mapped to

```yml
gehoertZuPlan:
  type: OBJECT
  sourcePath: [gehoertzuplan_pfs_plan_fk=_id]pfs_plan
  refType: pfs_plan
  properties:
    id: 
      type: INTEGER
      sourcePath: _id
    title: 
      type: STRING
      sourcePath: _id
    type: 
      type: STRING
      constantValue: pfs_plan
```

That means feature refs can be handled as plain objects or object arrays by the code, no special handling needed in most cases. When they need to be detected, that is still possible using `isFeatureRef`.

I carried over the the changes to `rel-as-key` from `feature-ref-object3`, the old behaviour is commented out and marked with `TODO` if you would like to re-enable it.

azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1091.surge.sh

cportele:
@azahnen
Thanks. I also did not like the introduction of `FEATURE_REF_OBJECT` in addition to `FEATURE_REF` and I also experimented with supporting the new requirements using `FEATURE_REF`. But that introduced a lot of other complexities. Good, if you found ways to avoid those. At least until now, `FEATURE_REF` is not an object, it is a `VALUE`, that is either a `STRING` or an `INTEGER`; both in the data and when used as a queryable or sortable. If the referenced feature is not in the same API, there is also no `refType`. I will have a look at the changes.

cportele:
@azahnen - I did tests with `strassen`. My observations:

### Without changes to the configuration:

* http://localhost:7080/rest/services/strassen/collections/abschnitteaeste/items/3713038A3713041O works as expected
* Using profiles `rel-as-key` and `rel-as-uri` in GeoJSON results in exceptions. I will have a look at those.
* http://localhost:7080/rest/services/strassen/collections/unfaelle/items/929 results in an exception: "Type VALUE may only be used with coalesce. Found: VALUE. Path: abs."

### With explicit properties 

in abschnitte

```yaml
      unfaelle:
        sourcePath: "[abs=abs]unfaelle_point"
        type: FEATURE_REF_ARRAY
        label: Unfälle
        description: Unfälle auf dem Abschnitt oder Ast
        refType: unfaelle
        properties:
          id: 
            type: INTEGER
            sourcePath: fid
          title: 
            type: STRING
            sourcePath: unfzeit
          type: 
            type: STRING
            constantValue: unfaelle            
```

In unfaelle:

```yaml
      abs:
        type: FEATURE_REF
        label: Abschnitt/Ast
        description: 16-stellige Kennung des Abschnittes oder Astes
        refType: abschnitteaeste
        properties:
          id: 
            type: INTEGER
            sourcePath: abs
          title: 
            type: STRING
            sourcePath: abs
          type: 
            type: STRING
            constantValue: abschnitteaeste            
```

* Requests to both features result in an exception.
  * For HTML: "Transformer OBJECT_REDUCE_FORMAT can only be applied to OBJECT or OBJECT_ARRAY, found: FEATURE_REF_ARRAY" 
  * For GeoJSON: "Transformer OBJECT_MAP_FORMAT can only be applied to OBJECT or OBJECT_ARRAY, found: FEATURE_REF_ARRAY"
  * The exceptions result in a 400 status, but this is not a client error, it should be 500 for an internal error. 
* If I add the `type` property, it is also not clear to me, why `refType` is still required. If I remove `refType`, I just get a normal object with `id`, `title`, and `type` properties. Shouldn't the type be determined from the `type` value, if it is provided?
* If I only change the configuration in  abschnitteaeste and leave unfaelle unchanged, requests to abschnitteaeste work again (with the correct title). I have not yet tried to understand why a change in the unfaelle schema results in a different response for abschnitteaeste features.

cportele:
@azahnen - Regarding the need for a variable type, there are basically two types how this could be expressed in the feature schema. The second option should already work in v3.5 (without the title).

1. Determine the `refType` from a column in a master table that contains all objects

```yaml
gehoertZuPlan:
  type: FEATURE_REF
  sourcePath: [gehoertzuplan_fk=_id]coretable
  properties:
    id: 
      type: INTEGER
      sourcePath: id
    title: 
      type: STRING
      sourcePath: id
    refType: 
      type: STRING
      sourcePath: objektart
      constraints: 
        enum: 
        - pfs_plan 
        - bra_ausbauplan
        - bst_netzplan
        - isa_plan
```

I have included the `refType` as a property of the reference object, but this can of course also be addressed differently. The enum constraint (or however this is eventually expressed) is optional, but relevant/useful for generating the schema for [OGC API Features Part 5: Schemas](https://docs.ogc.org/DRAFTS/23-058.html#rc_feature-references).

2. Based on the column used for the foreign key (using `concat` for an array or `coalesce` for a single reference)

Using `coalesce` and a simple link with no title:

```yaml
gehoertZuPlan:
  type: FEATURE_REF
  valueType: INTEGER
  coalesce:
    - sourcePath: gehoertzuplan_bra_ausbauplan_fk
      type: FEATURE_REF
      refType: bra_ausbauplan
    - sourcePath: gehoertzuplan_bst_netzplan_fk
      type: FEATURE_REF
      refType: bst_netzplan
    - ...
```

Or an array with `concat` and the link title from a feature property:

```yaml
hatObjekt:
  type: FEATURE_REF_ARRAY
  valueType: INTEGER
  concat:
    - sourcePath: "[id=gehoertzuplan_bst_netzplan_fk]bst_abwasserleitung"
      type: FEATURE_REF_ARRAY
      refType: bst_abwasserleitung
      properties:
        id:
          type: INTEGER
          sourcePath: id
        title: 
          type: STRING
          sourcePath: text
    - sourcePath: "[_id=gehoertzuplan_bst_netzplan_fk]bst_armatur"
      type: FEATURE_REF_ARRAY
      refType: bst_armatur
      properties:
        id:
          type: INTEGER
          sourcePath: id
        title: 
          type: STRING
          sourcePath: text
    - ...
```

azahnen:
@cportele
- The issues should be fixed. One issue is caused by the missing flattening implementation, `GeoJsonWriterProperties` assumes the upstream is flattened, but it isn't. To work around that, you have to disable the flattening in the `GEO_JSON` building block.
- The mapping with explicit properties should now work as expected. The `type` property has preference over `refType`, and `refType` is no longer needed if `type` is defined.
- Regarding dynamic types, yes, using an enum is what I would have expected.
- I restored the old behaviour in `SchemaDeriver`, but it does not handle the new cases yet. Could you take care of that? It should check for a `type` (`FeatureRefResolver.TYPE`) property with either a `constantValue` or an `enum`. Both would have preference over `refType`, which is set to `FeatureRefResolver.REF_TYPE_DYNAMIC` if not defined in the YAML.

cportele:
@azahnen - Thanks, I will update `SchemaDeriver`.

cportele:
@azahnen 

Some thoughts on the use of FEATURE_REF(_ARRAY) properties with a flatten transformation:

### The schema resource 

Path: `/collections/{collectionId}/schema`

If the provider schema includes FEATURE_REF(_ARRAY) properties and the GeoJSON module includes a flatten transformation, the schema resource returned by the API currently returns a schema with three properties "prop.id", "prop.title" and "prop.type" with no indication of a feature reference. 

This, however, is a bug - a leftover from the previous approach to schemas, where the schema was the GeoJSON schema. With the new OGC API Features Part 5 the GeoJSON transformations should not be taken into account when generating the schema resource. `SchemaCacheFeatures` has to be changed to remove the visitor that applies the GeoJSON transformations. 

That said, a flatten transformation can also be applied in the provider schema. There are two options:

1. Disallow that a FEATURE_REF(_ARRAY) property is used in a feature schema that includes a flatten transformation. Since those cases are rare - if there are any at all, this might be acceptable.
2. Do not flatten FEATURE_REF(_ARRAY) properties in the flatten transformation. This would result in the expected schema.

### Feature responses

Path: e.g., `/collections/{collectionId}/items` or `/search`

If a flatten transformation is applied for a negotiated format, the FEATURE_REF(_ARRAY) property is flattened and is no longer visible in the feature schema, only the properties of the FEATURE_REF object (id, title, type). The information about the FEATURE_REF property is lost. As a result, the handling the profile no longer works as expected.

Again, we have the two options:

1. Disallow that a FEATURE_REF(_ARRAY) property is used in a feature schema where a format is offered that supports the profile `rel-as-link` and includes a flatten transformation. That is, FEATURE_REF(_ARRAY) could no longer be used in the "strassen" API on demo.ldproxy.net or the flatten transformation would have to be removed for GeoJSON and JSON-FG.
2. Do not flatten FEATURE_REF(_ARRAY) properties in the flatten transformation. This will not yield the expected feature encoding, if a profile is negotiated that results in an object value for the property, since the value would not be flattend. For example, GeoJSON, which supports the `rel-as-link` profile. CSV, FlatGeobuf and MVT would still deliver the expected results since `rel-as-link` is not supported for those formats. Sub-options include:
  a. Disable formats like GeoJSON that support the profile `rel-as-link` and include a flatten transformation either for the format or in the provider. This would disable GeoJSON and JSON-FG some collections in the "strassen" API on demo.ldproxy.net. This is not an option and the flatten transformation would have to be removed for GeoJSON and JSON-FG.
  b. Remove `rel-as-link` from the supported profiles of those formats for those collections. This could be confusing, since the profile negotiation would differ from collection to collection for those formats and it would no longer allow that both the "href" and the "title" are included in flattened representations. On the other hand, where flattening is important, it might actually be the right choice to not support `rel-as-link` in GeoJSON, but use `rel-as-uri` as the default.
  c. Require that those formats resolve such cases in their encoder. However, this option would mean that flattening would have to be reimplemented in those formats, which should be avoided.

If we consider the multi-collection-query-capability offered by the Search module, the fact that different profiles may be applied depending on the collection in such a query seems to be a no-go. That is, in options 2a and 2b, the behavior should be the same for a format for all collections in the API.

### Conclusion

The simpler approach is options 1/1.

The best approach in terms of capabilities seems 2/2b.

### Impact on existing APIs

Currently the "strassen" API on demo.ldproxy.net is the only API that I know that uses both flattening and feature references. With option 1/1, the flatten transformations would have to be removed from "strassen" API on demo.ldproxy.net. Since this is a demonstration API, this should not be an issue.

Also with option 2/2b, the GeoJSON encoding in other APIs that currently use the explicit "Link" object types in the provider schema together with flattening (e.g., selected APIs on https://ogc-api.nrw.de/) would be affected, if the provider schema would be migrated to the FEATURE_REF types. This may or may not be an issue.

azahnen:
@cportele 
Thanks, sounds reasonable. For the schema resource, I would go with 1 for now. A third option might be to ignore the flattening transformation, even if it is defined in the provider. Or fourth, disallow flattening in the provider.

For feature responses, would a flattened link object be a desired/acceptable result for `rel-as-link`? That should be reachable by tweaking the transformation order, i.e. the profile transformation would have to be applied before the flattening transformation.

cportele:
@azahnen 

> A third option might be to ignore the flattening transformation, even if it is defined in the provider. Or fourth, disallow flattening in the provider.

Yes. Ignoring a flattening transformation in the provider would be the correct approach, if we would consider the flattened and unflattened representation of the same feature as different profiles, which conceptually would be the correct approach. That is, the Part 5 feature schema would be based on the unflattened properties. The difference to the `rel-as-xxx` profiles is that we currently don't support that the client can request the profile, it is determined by the configuration. 

In that sense, it would seem reasonable to disallow flattening in the provider and restrict flattening to the API building blocks.

Just thinking out loud: To use the "profile" parameter also for flattening, the following would be needed:

- New profiles, e.g., `flatten` and `regular`.
- The default negotiation would be `regular` unless `flatten` is requested.
- Formats that only support the flattened representation (CSV, FlatGeobuf, etc.) would always negotiate to `flatten`.
- Formats the support both could have an option to change the default for an API to `flatten`, that is, `regular` would only be applied, if explicitly requested. This option would be enabled where a configuration currently includes a flatten transformation for a format.
- All flatten transformations could be removed from the API configurations.

The `flatten` profile could also be standardized in the future, if there is interest. It would then no longer be an ldproxy-specific internal configuration option.

> For feature responses, would a flattened link object be a desired/acceptable result for rel-as-link? 

I was pondering this question myself when I did the analysis. In a way, the answer is yes, because this is what we did with the Link object also in flattened configurations. So, if that could be supported by the transformation order, then that would be best.

azahnen:
@cportele 
Concat and coalesce are ready for testing.

I am still working on the flattening, I would then try to go with 4/3, i.e. disallow flattening in the provider and apply profile transformations before flattening.

The `flatten` and `regular` profiles sound interesting, I think that should be the way forward. The functionality is very similar to the rel-profiles and that would give even more control to the clients. I think the main reason why the NRW-APIs are flattened is because that is what their own/internal clients need. But other clients of the public APIs might prefer the regular encoding, and with the profiles we could support both.

cportele:
@azahnen - Thanks, I will start testing. +1 on the other aspects.

cportele:
@azahnen 

I am running into some problems in my tests:

1. In the "coretable" configuration (branch "transformation-pipeline") a request to http://localhost:7080/rest/services/xplangml_6_0/collections/bp_baugrenze/items throws an exception at the `checkState()` method at `SchemaMapping.java:73`. 
2. A request to http://localhost:7080/rest/services/xplangml_6_0/collections/bp_plan/items throws an IndexOutOfBoundsException exception at `FeatureTokenReader.java:225`.
3. When you enable the queryables in the services configuration, there are numerous errors like "Query parameter for property 'wirdDargestelltDurch.0_id' at path '/collections/bp_anpflanzungbindungerhaltung/items' could not be created, the property was not found in the feature schema". Properties with `concat` or `coalesce` should not be queryables, but the detection in `OnlyQueryables` no longer works, probably because the internal logic in the transformation pipeline has changed.
4. In general, queryables/sortables on FEATURE_REF properties no longer seem to work. I think this is due to the change from value (string/integer) to object for FEATURE_REF properties. The FEATURE_REF property as a queryable has to continue to evaluate to the key (in the simple cases with a refType just the id as it is now). A conceptual question is what we do for cases where the type varies. Either we do not support queryables those cases or use the keyTemplate to filter on both the id and type properties (which would have to be columns in the same table).

azahnen:
@cportele 
Flattening is ready for testing.

I will look into 1. and 2. tomorrow.

Regarding 3., can properties of a concat/coalesce object_array/object be queryable? If not, the if-block starting in line 55 of `OnlyQueryables` can be deleted  and the parent schemas should be checked for concat/coalesce in line 48. 

Regarding 4., I will fix the simple case tomorrow. Dynamic types cannot be supported as of now, the solution you propose would require that the keyTemplate is resolved as part of the SQL query, which is not the case.

cportele:
@azahnen 

Thanks. 

Re 3: No, properties with concat or coalesce are not eligible to be queryable or sortable.

Re 4: OK, I think we can restrict queryables and sortables to the simple case. If we would also want to support the dynamic type case, we would have to parse the value based on the key template to determine the type and id values and query on both values with an AND. In principle that should be possible, but may be quite a bit of work to get it right, so we should not do this now.

azahnen:
@cportele 
I am still working on issues with coretable. The new feature refs did not work at all with subdecoders, I had to implement some additional transformers for that. 
I had it mostly working when I realized that I used the main branch of coretable. With the transformation-pipeline branch some new issues popped up. And it looks like feature refs now use plain sql, so my work on subdecoders might not have been necessary.

cportele:
@azahnen - Thanks for the update. The sub-decoder support will also be necessary, some references may still be in the JSON.

azahnen:
@cportele 
Please test again, there should be no more errors. I also pushed some small fixes to `coretable`.

I will now look into 4.

cportele:
Thanks @azahnen, I will test it.

I have also made tests with an updated, ShapeChange-generated configuration. A small number of collections do not start, because they have no ID property. All these types reference a schema '#/fragments/BP_Flaechenschlussobjekt'. Every property from BP_Flaechenschlussobjekt or its supertypes is not included in these cases. However, BP_Flaechenschlussobjekt is also used in other types and there no issues are reported. I don't think this is related to the level of nesting - I think you had increased the limits quite generously and a warning would be written, if a limit would stop the schema resolution, correct? 

I have created a new branch `sc` for this "coretable" configuration.

```
ldproxy_coretable  | INFO  [2023-11-22 18:13:56,969]             xplangml_6_0 - Feature provider with id 'xplangml_6_0' started successfully. (min connections=1, max connections=150, stream capacity=10)  
ldproxy_coretable  | ERROR [2023-11-22 18:13:57,943]             xplangml_6_0 - Collection 'BP_BaugebietsTeilFlaeche' has been disabled because its feature type has no property with role ID in the provider schema.  
ldproxy_coretable  | ERROR [2023-11-22 18:13:57,944]             xplangml_6_0 - Collection 'BP_BesondererNutzungszweckFlaeche' has been disabled because its feature type has no property with role ID in the provider schema.  
ldproxy_coretable  | ERROR [2023-11-22 18:13:57,944]             xplangml_6_0 - Collection 'BP_GemeinbedarfsFlaeche' has been disabled because its feature type has no property with role ID in the provider schema.  
ldproxy_coretable  | ERROR [2023-11-22 18:13:57,944]             xplangml_6_0 - Collection 'BP_GruenFlaeche' has been disabled because its feature type has no property with role ID in the provider schema.  
ldproxy_coretable  | ERROR [2023-11-22 18:13:57,945]             xplangml_6_0 - Collection 'BP_SpielSportanlagenFlaeche' has been disabled because its feature type has no property with role ID in the provider schema.  
ldproxy_coretable  | ERROR [2023-11-22 18:13:57,945]             xplangml_6_0 - Collection 'BP_UeberbaubareGrundstuecksFlaeche' has been disabled because its feature type has no property with role ID in the provider schema.  
ldproxy_coretable  | ERROR [2023-11-22 18:13:57,945]             xplangml_6_0 - Collection 'BP_VerEntsorgung' has been disabled because its feature type has no property with role ID in the provider schema.  
ldproxy_coretable  | ERROR [2023-11-22 18:13:57,945]             xplangml_6_0 - Collection 'BP_WohngebaeudeFlaeche' has been disabled because its feature type has no property with role ID in the provider schema.  
ldproxy_coretable  | ERROR [2023-11-22 18:13:57,949]             xplangml_6_0 - Collection 'SO_Strassenverkehr' has been disabled because its feature type has no property with role ID in the provider schema.  
```

azahnen:
@cportele 
I pushed a fix. I am pretty sure `BP_Flaechenschlussobjekt` is only used in the 9 types that throw an error, so the behaviour was consistent. It was a bug in the fragment resolver, when a fragment referenced from a merge contained another schema reference, that reference was ignored.

cportele:
@azahnen - Thanks, I confirm it now works as expected also for me, too. 

`BP_Flaechenschlussobjekt` is also referenced in `BP_FlaecheOhneFestsetzung` and `BP_KleintierhaltungFlaeche`, but without properties and no other schema.

cportele:
@azahnen - The demo configuration is in branch `transformation-pipeline`.

I also noticed that queryables are handled, but not yet sortables.

cportele:
@azahnen 

With requests to `items` on `demo/cologne_lod2` collection `building` and `geoval/geoval` collection `explorationsite` I get "No value present" at `FeatureTokenTransformerRemoveEmptyOptionals.java:68` (`context.schema()` is `empty`).

azahnen:
@cportele 
I pushed a fix for the queryables issue. This works as expected now:

http://localhost:7080/rest/services/strassen/collections/unfaelle/items?f=json&abs=4511098O4511033A

cportele:
@azahnen - It works for me, too.

However, http://localhost:7080/rest/services/strassen/collections/abschnitteaeste/items?filter=a_contains(unfaelle,[1]), results in "Filter is invalid. Unknown property: unfaelle".

cportele:
@azahnen - Another case you have to check. Configuration: `xleit`. Request: http://localhost:7080/rest/services/xtrassePFS/collections/pfs_plan/items/1?f=json

The "hatObjekt" and "hatPFSObjekt" values are not correct. It looks as if there are three values per ref, one with the title, one with the type and one with the id:

```json
      {
        "title": "",
        "href": "http://localhost:7080/rest/services/xtrassePFS/collections/bst_speicher/items/"
      },
      {
        "title": "",
        "href": "http://localhost:7080/rest/services/xtrassePFS/collections//items/2"
      },
      {
        "title": "Speicher Jemgum",
        "href": "http://localhost:7080/rest/services/xtrassePFS/collections//items/"
      }
```

With `profile=rel-as-key` the properties are missing.

azahnen:
@cportele `building` should be fixed.

cportele:
@azahnen - The behavior is unchanged for me with `demo/cologne_lod2`, collection `building`.  http://localhost:7080/rest/services/cologne_lod2/collections/building/items still results in:

```
...
ldproxy_demo  | ERROR [2023-11-23 14:43:27,184]             cologne_lod2 - No value present  
ldproxy_demo  | DEBUG [2023-11-23 14:43:27,184]             cologne_lod2 - Stacktrace:  
ldproxy_demo  | java.util.NoSuchElementException: No value present
ldproxy_demo  | 	at java.base/java.util.Optional.get(Unknown Source)
ldproxy_demo  | 	at de.ii.xtraplatform.features@6.4.0-transformers-SNAPSHOT/de.ii.xtraplatform.features.domain.FeatureTokenTransformerRemoveEmptyOptionals.onArrayStart(FeatureTokenTransformerRemoveEmptyOptionals.java:68)
ldproxy_demo  | 	at de.ii.xtraplatform.features@6.4.0-transformers-SNAPSHOT/de.ii.xtraplatform.features.domain.FeatureTokenTransformerBase.onArrayStart(FeatureTokenTransformerBase.java:148)
ldproxy_demo  | 	at de.ii.xtraplatform.features@6.4.0-transformers-SNAPSHOT/de.ii.xtraplatform.features.domain.FeatureTokenReader.emitEvent(FeatureTokenReader.java:196)
ldproxy_demo  | 	at de.ii.xtraplatform.features@6.4.0-transformers-SNAPSHOT/de.ii.xtraplatform.features.domain.FeatureTokenReader.onToken(FeatureTokenReader.java:47)
...
```

azahnen:
@cportele `pfs_plan` and `building` should be fixed. For CityJSON there seems to be an error also in master:

https://demo.ldproxy.net/cologne_lod2/collections/building/items?f=cityjson

cportele:
@azahnen - `xleit`, `cologne_lod2` as well as `geoval` now also work for me. 

I will take care of the CityJSON issue. The FEATURE_REF encoding in GML is also not correct yet, I will take care of thatt, too.

azahnen:
@cportele 
Queryables and sortables should now work as expected. I implemented a cleaner approach that adds an additional property for filters and sorting. So given for example this

```yaml
hatObjekt:
  sourcePath: '[_id=gehoertzuplan_pfs_plan_fk]bst_abwasserleitung'
  type: FEATURE_REF_ARRAY
  refType: bst_abwasserleitung
  properties:
    id:
      sourcePath: _id
      type: INTEGER
    title:
      sourcePath: text
      type: STRING
```

the `FeatureRefResolver` adds this

```yaml
hatObjekt_queryable:
  sourcePath: '[_id=gehoertzuplan_pfs_plan_fk]bst_abwasserleitung/_id'
  type: VALUE_ARRAY
  valueType: INTEGER
  excludedScopes: [RETURNABLE,RECEIVABLE]
  transformations:
    rename: hatObjekt
```

The `rename` transformation is then applied in the appropriate places, so that the additional property replaces the original one in queryables and sortables schemas.
The tricky part was that the SQL query generation now has to use different `SchemaSql` instances for generating columns and resolving filters and sort keys.

cportele:
@azahnen - I will record any issues that I discover in this comment.

- [ ] When I request features in CSV or FlatGeobuf, an exception occurs in `FeatureTokenValidator.onArrayStart()`, because the elements in the path stack are "geometry", but the context path is "geometry[1]", which fails the `doesNotStartWithPreviousPath` test. When I request the features as GeoJSON, the context path is "geometry" and the features are returned. Example: http://localhost:7080/rest/services/daraa/collections/CulturePnt/items?f=csv

azahnen:
@cportele Should be fixed.

cportele:
@azahnen - Indeed, thanks for the quick fix.

A question: A property with `coalesce` (example: `wert` in `XP_GenerAttribut` in `coretable`) is never an array, but in the schema it is converted to an array (as far as I can see through `MappingOperationResolver.resolveCoalesce()`). As a result, `wert` is of type `VALUE_ARRAY`. I assume this is because technically there are multiple values, we just pick the first one. Is that correct? 

If yes, `SchemaDeriver` needs to be adjusted, because it currently expects type `VALUE` with `coalesce` (as it is in the provider schema). I can do the changes.

azahnen:
@cportele Yes, but that is reversed by the `coalesce` transformer. If you apply `WithTransformationsApplied` without any `additionalTransformations`, you should get the correct schema.

cportele:
@azahnen - Ah ok, I will try this out...

cportele:
@azahnen - Yes, that works!

Another related question: When the value of `wert` is passed through the pipeline, there does not seem to be information in the context from which coalesce option the value originates. As a result, the value is always treated as a string, because the `valueType` is `VALUE`. Should the `valueType` be updated in the coalesce transformer when the value is determined? 

```yaml
      wert:
        type: VALUE
        label: wert
        coalesce:
        - sourcePath: wert_double
          type: FLOAT
        - sourcePath: wert_integer
          type: INTEGER
        - sourcePath: wert_string
          type: STRING
```

azahnen:
@cportele Yes, in theory `context.valueType()` should return the correct type. But the coalesce transformer cannot set it because the source paths are no longer available. It has to happen one step earlier when the buffer is written. When the buffer has the correct types and coalesce chooses one, then `context.valueType()` should be correct.

I will take a look.

cportele:
@azahnen - Thanks. 

There is another issue related the schema derivation. When I apply the `WithTransformationsApplied()` visitor, the type is indeed back as `VALUE`, but all coalesce items now have type `VALUE` and valueType `STRING`, also for those cases where the value type originally was `INTEGER` or `FLOAT`.

cportele:
@azahnen - Forget the last issue, in the coalesce items `type` was set, not `valueType`.

cportele:
@azahnen

> The `rename` transformation is then applied in the appropriate places, so that the additional property replaces the original one in queryables and sortables schemas.

Since we have to apply `WithTransformationsApplied` also before deriving the returnables/receivables schema (see above), this has the effect that queryable FEATURE_REF properties are not included in the schemas, because in the additional property the returnable/receivable scopes are excluded.

azahnen:
@cportele 
Coalesce should be fixed.

Regarding the transformations, would it be sufficient to pass an optional scope to `WithTransformationsApplied` and then only execute transformations for properties that match the scope?

azahnen:
Or event simpler, apply `WithScope` before `WithTransformationsApplied `?

cportele:
@azahnen 

Ah yes, that should work. Right now, I have applied `WithScope` after `WithTransformationsApplied`, but reversing this should to the trick.

cportele:
@azahnen - There is again an issue when requesting buildings in `cologne_lod2` (at least for formats HTML and JSON-FG). The trace shows that a geometry object is opened, but not closed before the next geometry object is opened.

```
TRACE [2023-11-26 14:29:00,685]                          - START FEATURE  []  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,697]                          - VALUE gml_id DENW37AL1000qsEJ  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,700]                          - START OBJECT bbox []  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,705]                          - END OBJECT bbox []  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,707]                          - VALUE creationDate 2023-04-07  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,707]                          - VALUE measuredHeight 21.766  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,707]                          - VALUE storeysAboveGround 4  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,717]                          - VALUE roofType Satteldach  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,717]                          - VALUE function Wohnhaus  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,717]                          - VALUE datenquelleBodenhoehe Verschneidung mit DGM1  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,717]                          - VALUE datenquelleDachhoehe Laser-Scan  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,718]                          - VALUE datenquelleLage Liegenschaftskataster  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,718]                          - VALUE gemeindeschluessel 05315000  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,718]                          - VALUE grundrissaktualitaet 2021-10-01  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,718]                          - VALUE geometrietyp2dReferenz 3000  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,718]                          - START OBJECT address [1]  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,719]                          - START OBJECT address.multiPoint [1]  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,719]                          - END OBJECT address.multiPoint [1]  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,719]                          - VALUE address.ThoroughfareName Weyerstr.  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,719]                          - VALUE address.ThoroughfareNumber 49  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,719]                          - VALUE address.LocalityName Köln  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,719]                          - VALUE address.CountryName Germany  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,719]                          - END OBJECT address [1]  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,719]                          - END ARRAY address  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,719]                          - START OBJECT lod2Solid []  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,723]                          - END OBJECT lod2Solid []  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,723]                          - START OBJECT surfaces.lod2MultiSurface [1]  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
TRACE [2023-11-26 14:29:00,723]                          - START OBJECT surfaces.bbox [1]  [d.i.x.features.domain.FeatureTokenValidator] [pool-23-thread-60]  
ERROR [2023-11-26 14:29:00,726]             cologne_lod2 - Cannot open object with path [surfaces, bbox]. Current nesting stack:
[surfaces, lod2MultiSurface] O
 [c1819244-1298-48c5-8d7a-b07df6d70486] [d.i.x.features.domain.FeatureStream] [dw-34]  
DEBUG [2023-11-26 14:29:00,726]             cologne_lod2 - Stacktrace: [c1819244-1298-48c5-8d7a-b07df6d70486] [d.i.x.features.domain.FeatureStream] [dw-34] #STACKTRACE 
java.lang.IllegalStateException: Cannot open object with path [surfaces, bbox]. Current nesting stack:
[surfaces, lod2MultiSurface] O

	at de.ii.xtraplatform.features@6.4.0-SNAPSHOT/de.ii.xtraplatform.features.domain.FeatureTokenValidator.error(FeatureTokenValidator.java:125)
	at de.ii.xtraplatform.features@6.4.0-SNAPSHOT/de.ii.xtraplatform.features.domain.FeatureTokenValidator.onObjectStart(FeatureTokenValidator.java:43)
```

azahnen:
@cportele Should work again. `surfaces` should be removed in HTML and JSON-FG collections, but the transformer was not applied anymore since I cleaned up a bit too much.

cportele:
@azahnen - It works for me now with JSON-FG, but in HTML I get a new NPE:

```
java.lang.NullPointerException: Cannot invoke "de.ii.xtraplatform.features.domain.SchemaMapping.getSchemasByTargetPath()" because "schemaMapping" is null
	at de.ii.xtraplatform.features@6.4.0-SNAPSHOT/de.ii.xtraplatform.features.domain.transform.TransformerChain.explodeWildcard(TransformerChain.java:76)
```

azahnen:
@cportele Fixed.

cportele:
@azahnen - Thanks, now it works for me, too.

cportele:
@azahnen - yet another NPE: this time in `demo/cologne_lod2` when I delete the 3D Tiles cache and seeding starts.

```
java.lang.NullPointerException: Cannot invoke "java.util.Map.entrySet()" because "entries" is null
	at de.ii.xtraplatform.features@6.4.0-SNAPSHOT/de.ii.xtraplatform.features.domain.ModifiableGenericContext.setMappings(ModifiableGenericContext.java:545)
	at de.ii.xtraplatform.features@6.4.0-SNAPSHOT/de.ii.xtraplatform.features.domain.ModifiableGenericContext.setMappings(ModifiableGenericContext.java:24)
	at de.ii.xtraplatform.features.sql@6.4.0-SNAPSHOT/de.ii.xtraplatform.features.sql.app.FeatureDecoderSql.init(FeatureDecoderSql.java:86)
	at de.ii.xtraplatform.features@6.4.0-SNAPSHOT/de.ii.xtraplatform.features.domain.FeatureTokenTransformerBase.init(FeatureTokenTransformerBase.java:110)
        ...
```

azahnen:
@cportele The first one should be fixed.

azahnen:
@cportele The second one should be fixed as well.

cportele:
@azahnen - Yes, both as fixed 👍

cportele (approved):
I have a small list of open issues that we can address in follow-up pull requests.

PR #1102, opened 2023-11-24
relax requirement for a persistent ID for building parts

cportele:
### Pull request checklist

-   [x] Updated documentation

Currently the CityJSON module requires that every building part has a unique "id" property. If not, an exception is thrown when a building part is encountered. Strictly, it is not necessary that the id must be persistent across requests. From an API perspective, it is sufficient, if the building ids are persistent. This change generates a UUID as the identified, if no id is provided for a building part, instead of throwing an error.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1102.surge.sh

azahnen (approved):

PR #1126, opened 2023-12-19
remove support for custom style metadata

### LINKED ISSUE #1115: remove support for custom style metadata
This will remove support for custom style metadata that is defined in`.metadata` files. That means all metadata will be automatically derived.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1115.

If custom metadata will be supported in the future, it would be embedded in the MapLibre or 3D Tiles stylesheet.


azahnen (approved):
Can the dependency on `com.github.java-json-tools:json-patch` be removed or it is used for something else?

cportele:
@azahnen - It can be deleted. I was confused by differences between the Java path and the name of the dependency and assumed that the `com.github.fge.jsonpatch` package was also used elsewhere. I will remove the dependency and then merge.

PR #1132, opened 2023-12-31
avoid memory leaks on entity reload

### LINKED ISSUE #1088: Significant increase in memory consumption after reloading configs via tasks/reload-entities
### Discussed in https://github.com/interactive-instruments/ldproxy/discussions/1063

<div type='discussions-op-text'>

<sup>Originally posted by **schmity** October  2, 2023</sup>
We use tasks/reload-entities?ids=*&types=providers in the Admin-connector to reload new or changed ldproxy-configs. Usually configs change or few are added.

Monitoring ldproxy memory consumption, we observe it increases significantly on initial reload after startup.
In an instance with 336 APIs memory consumption increases from 1.4 GB to 2,3 GB.

Further calls to reload-entities still increases memory consumption, but much less (several MB per call).

Given that configs only changed or a few are added, memory consumption should not change significantly after reload via reload-entities.

From our logs: 

requests.log
NO_IP_LOGGING - - [29/Sep/2023:15:36:00 +0200] "POST /ldproxy_intranet/reload-entities?ids=*&types=providers HTTP/1.1" 200 - "-" "axios/0.26.1" X-Forwarded-For: "-" "-" "-" 1835262 22876633249664 37159690
NO_IP_LOGGING - - [29/Sep/2023:15:36:00 +0200] "POST /ldproxy_intranet/reload-entities?ids=stadtgruen&types=services HTTP/1.1" 200 - "-" "axios/0.26.1" X-Forwarded-For: "-" "-" "-" 3919716 22876633249664 254901

procmeminfo.log monitoring memory consumption
Fri 29 Sep 2023 03:35:01 PM CEST ldproxy_intranet PID: 1665356 Resident Memory used: (RSS) 1371756 kB
Fri 29 Sep 2023 03:40:01 PM CEST ldproxy_intranet PID: 1665356 Resident Memory used: (RSS) 2284980 kB 
Fri 29 Sep 2023 03:45:01 PM CEST ldproxy_intranet PID: 1665356 Resident Memory used: (RSS) 2285236 kB</div>

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1088

Depends on:

- https://github.com/interactive-instruments/xtraplatform/pull/146
- https://github.com/interactive-instruments/xtraplatform-spatial/pull/251

Environment used for testing: 

- Base configuration: ri.ldproxy.net (daraa and vineyards reduced to Features Part 1 and 2)
- Change: With 500 additional daraa and vineyards APIs, so in total 501 daraa and 501 vineyards APIs.

There were several issues:

1. During entity reload, the new data object still has to be set in the entity and replace the old data, even if there was no change. Otherwise there can be side-effects, for example, the new data object may already be in a cache in the entity data store (`EntityDataStoreImpl.eventSourcing.cache`) which effectively keeps both the old and new data object in memory.
2. When an API entity is reloaded or removed, the existing change listeners (from Feature Core and Tile Seeding) are still registered in the handler. They need to be removed on entity reload. 
3. Old API definitions of endpoints remain in the cache when a service entity is reloaded. They need to be removed from the API definition cache on reload.
4. References to objects that have to be removed during reload must also be removed during shutdown of an entity.
5. Unnecessary reloads are executed when `createdAt` and `lastModified` are not set, because this results in different entity data hash values even if there is no effective change in the entity data. To avoid this, both properties are marked as auxiliary.

There are two other reasons why memory increases as a result of a reload in the test setup:

- The change to address the first issue has a side-effect that increases the memory, too, but to a far lesser extent. Some entities store information from the entity data also in other objects during `onStartup()`. For example, `FeatureProviderSql` stores `SqlPathDefault` or `EpsgCrs` objects as part of their variables. On a reload event of unchanged entity data, the entity does not change - except for its data object, which now has fresh instances of those objects. In my tests this only affected smaller objects, so this limited duplication may be acceptable, in particular since repeated reloads do not increase the instance count.
- The event queue in `EventStream` grows. In a test with the standard `ri` configuration (daraa provider/service, vineyards provider/service, service defaults) each reload added 11 new ReplayEvent instances (which include the payload, which can be large for complex providers or services). This can become a memory issue, if an instance runs for a long period with frequent reload events. This should be addressed as part of the cloud native epic.

I also did a test with a configuration that includes tiles and styles:

- There is a potential issue with reloading tile providers that is independent of the memory issue. In the tile providers, in particular `TileProviderFeatures`, several variables are initialized in the constructor with information from the data object. This has the issue with redundant copies of some objects on a reload with unchanged data described above (but in this case with a larger memory footprint, because the complete data object is stored). But the main issue is that if the configuration has changes this may result in errors, because the `tileGenerator` or the `tileEncoders` are not updated on a reload but still have a copy of the initial data object.
- The memory increases on the first "reload" (data objects were unchanged only the data is set, no reload occurs) with a larger number of additional `FeatureSchema` objects, but only on the first reload. I have not been able to identify the reason for the increase, but there are also new `ImmutableBuildableMap` and `...$Builder` objects.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1132.surge.sh

cportele:
One other observation: There are two pauses of 50ms during every entity reload. The sleep time during the reload in the test setup took 200s (2006 entities) even though no entity was eventually reloaded. The CPU time of the reload was in the order of 10s. 

This is probably not an issue, since this is an artificial test setup and in practice one will reload particular entities that are known to have changed.

PR #1135, opened 2024-01-02
Fix tiles

### LINKED ISSUE #1134: Tiles: Issues in the edge case that tiles are only enabled for selected collections
While testing the fix for issue #1088, a few issues with tiles were detected that occur in the edge case that tiles are enabled only for (selected) collections, not for the dataset.

- If the store contained MBTiles files for disabled collections, these were enabled, too.
- No format was set in the schema of the `f` parameter.
- No format was set in the response to the tile request for a collection.
- All collections were included in the schema of the `collectionId` path parameter, including collections not listed in the tile provider.
- An exception was thrown when trying to determine an operation id on the dataset level.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1134.

Depends on:

- https://github.com/interactive-instruments/xtraplatform-spatial/pull/253


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1135.surge.sh

PR #1140, opened 2024-01-04
ignore feature references with id=null

### LINKED ISSUE #1139: Feature reference included on null value
A feature reference property is included, even if there is no reference and the `id` value (the foreign key) is `null` when the source path to the `id` property is a column in the feature table. 

The result are links like:

```json
"foo": {
  "title": "",
  "href": "http://localhost:7080/rest/services/bar/collections/baz/items/"
}
```

If the source path includes a join or if the feature reference is always present, the feature references are handled correctly.

If the `id` is `null`, the feature reference object must be removed entirely in the pipeline.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1139

Depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/254


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1140.surge.sh

azahnen (approved):

PR #1156, opened 2024-01-25
Various seeding issues

### LINKED ISSUE #1155: Tiles: various seeding issues
Several issues were observed in complex seeding tasks:

- Sometimes, JTS throws a topology error when trying to cluster adjacent/overlapping features. In those cases, the feature under test should simply be ignored for the cluster.
- Sometimes, invalid geometries are generated. In those cases, try to make the geometry valid again. 
- Purging tiles only deletes tiles in the bbox of the feature type, but there may exist additional empty tiles. All tiles in the tile matrix should be deleted on purge.
- MBTiles: When a large number of tiles are deleted, the query to delete the tile blobs can become larger than the limit for SQLite queries. In those cases, the deletes should be in batches.
- MBTiles: An error when seeding a tile stops the seeding process. It would be better to just log the error and continue seeding.
- MBTiles: Very rarely an error occurs when writing a tile (table tile_map does not exist, database corrupt), however, after that writing tiles works again without issues. On an error additional write attempts should be made before reporting a failed write attempt.
- MBTiles: The size of the cache always grows since `vacuum` is never executed on the MBTiles files. After deletes (purging, multi-level cleanup) a `vacuum` query should be executed on each MBTiles file in a store.

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1155.

Depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/260.

### DIFF
--- a/ldproxy-cfg/src/main/java/de/ii/ldproxy/cfg/Builders.java
+++ b/ldproxy-cfg/src/main/java/de/ii/ldproxy/cfg/Builders.java
@@ -13,6 +13,10 @@ public interface Builders {
     return new EntityDataBuilders() {};
   }
 
+  default ValueBuilders value() {
+    return new ValueBuilders() {};
+  }
+
   default OgcApiExtensionBuilders ogcApiExtension() {
     return new OgcApiExtensionBuilders() {};
   }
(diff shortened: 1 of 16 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1156.surge.sh

azahnen (approved):

PR #1174, opened 2024-03-08
Cloud native

### LINKED ISSUE #814: optimize startup performance
During every startup some computations and generations take place, which can become really slow for big datasets and/or many APIs. 
The goal of this issue would be to reduce the amount of times these computations are executed to a minimum by introducing caching mechanisms or similar means and only execute them when the source values have changed.
A special case of this are multi-node setups where every node executes the same computations at startup. So any solution needs to support sharing of the computed values.

### LINKED ISSUE #879: define dependencies and states for entities and building blocks
Identify all external dependencies and define how a component should behave when certain dependencies are missing. 
To communicate restrictions due to missing dependencies to consumers, every component should have one or more states. There are already coarse-grained states for entities which would allow to mark an entity as defective, but more fine-grained and context-specific states might be needed.

##### Feature Provider

A PGIS feature provider needs a PostgreSQL database, if it is unavailable the provider would be marked as defective. But not every functionality of the provider needs the database. 
The provider functionality is already partitioned into several interfaces with related `supportsXYZ` methods which could be leveraged for this. So for example the interfaces `FeatureCrs` and `FeatureExtents` might still be usable without a database (if extents are cached).  `FeatureQueries` and `FeatureTransactions` will not be supported in this state.

##### Building Block Features Core

This building block depends on one or more feature providers. If the feature provider currently cannot handle queries like described above, the building block would return 503 for all requests to `/items` and `/items/{id}`. 
The building block also provides the collection metadata for `/collections` and `/collections/{id}`, which would still be possible in the outlined scenario. But this metadata normally contains links to `/items`, which is unavailable. So the links could either be removed or e.g. the title could be changed to contain something like `(temporarily unavailable)`.

##### Tile Provider

Similarly a tile provider might need an S3 tile cache. If it is not available, tiles can neither be read from nor written to the cache. But it might still be possible to provide dynamically generated tiles and metadata. 
So if all tiles can be dynamically generated and the cache is only used for performance, the provider might either hide this fact from consumers and remain in the `active` state or it might switch into a `slow` state.
If tiles cannot be dynamically generated, it would switch into a `defective` state, but it could still provide metadata similar to the feature provider example.

##### Building Block Tiles

If the tile provider outlined above is in the `slow` state, this information might be added to the HTML view.
If the tile provider can only provide metadata, all requests to actual tiles would return 503. The metadata contains a URL template to access tiles, which might again be marked with `(temporarily unavailable)`. The HTLM view normally contains a map, which would be replaced with a similar message.
There are also other building blocks that depend on tiles like styles, which adds a web map link to the landing page. This link would also need to be removed or adjusted, which means the tiles building block itself also must communicate its restricted state.

### LINKED ISSUE #880: make entity startup async and event-driven
Currently entities (and building blocks) expect all external dependencies (e.g. other entities or databases or resources) to be available on startup. To ensure that, startup runs synchronously in a single thread, with only one entity at a time and entities ordered by type (codelists < providers < services). The `onStartup` method is then used to run all kinds of initialization, for example spatial extent computation.

Building on #879, this will replace this rigid system with a more flexible and resilient approach. Any initialization that requires external dependencies would no longer happen on startup. Instead, the dependent initialization would be triggered by a state change of the dependency.

That means startup itself will be a lot faster, but entities and building blocks will at first be in their most basic, possibly non-functional state. Most initialization would then happen asynchronously, when the dependencies arrive, which also means it could run in parallel. 

There might be an option to mimic the old behavior.

##### Before

If the required PostgreSQL database is not available, a feature provider will not start. If the feature provider is not available, a related service will not start. 
If the database becomes available, the feature provider and service remain stopped. A manual restart of the application is needed to reach the desired state of a running service.
If the database vanishes again after successful startup, the feature provider and service remain started. Instead of communicating this restriction to consumers/clients, they will throw exceptions on related operations.

##### After

Despite the PostgreSQL database not being available, both the feature provider and the service start. Of course the service respectively some of its building blocks will be in a restricted state and communicate this state appropriately to consumers/clients, see #879. 
The feature provider will check periodically for database availability. When it arrives, the provider will run any dependent initialization and change its state to `healthy`. The service, which has subscribed to state changes of the provider, will also run any related initialization and change its state to `healthy`.
If the database vanishes again, the provider will change its state to e.g. `defective`, which will again trigger the service and/or the building blocks to go back to the basic state.

### LINKED ISSUE #881: add health checks for external dependencies
As a follow-up to #879, it should be as easy as possible for an administrator to check the current state of entities and spot any missing external dependencies. 
ldproxy already has support for health checks which can be observed on the operations port (http://localhost:7081/healthcheck?pretty=true). This would add health checks for all the dependencies and states identified in #879.

### LINKED ISSUE #1114: remove deprecated store layout
This will remove the old store layout described in the docs in [Store](https://v3.docs.ldproxy.net/application/40-store.html). Only the new layout described in [Store (new)](https://v3.docs.ldproxy.net/application/41-store-new.html) will continue to be supported.

### LINKED ISSUE #1117: support reloading of values
This is a follow-up to #1077. 
Since codelists were entities before, they could be reloaded using `xtractl entity reload` or `watch: true`. The resources like MapLibre styles were never cached in memory, so they could just be changed on disk. Now an application restart is needed to reload any values.

This will add reload support for values in the same way as it is for entities.

Also see interactive-instruments/xtraplatform-cli#3.

### LINKED ISSUE #1149: configurable minimum maturity and maintenance levels
Add an option in `cfg.yml` to ignore modules that don't match the minimum levels. The default would be:

```yml
modules:
  minMaturity: CANDIDATE
  minMaintenance: LOW
```

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #814.

A persistent cache was introduced that is used for feature extents and counts in the SQL provider. By default, the cache is invalidated when the configuration changes or on feature/dataset change events. To recompute on startup like before, see [assumeExternalChanges](https://interactive-instruments-ldproxy-pr-1174.surge.sh/providers/feature/10-sql.html#connection-info).

Closes #879.
Closes #880.

Async startup can be disabled, it should then behave like before, see [startup](https://interactive-instruments-ldproxy-pr-1174.surge.sh/application/20-configuration/80-modules.html).

Closes #881.
Closes #1114.
Closes #1117.
Closes #1149.

Note that because of the chosen defaults many configurations might behave differently. To change the minimum levels, see [modules](https://interactive-instruments-ldproxy-pr-1174.surge.sh/application/20-configuration/80-modules.html). (I thought about using `PROPOSAL` and `NONE` as defaults in the dev environment, but decided against it because it might be confusing.)
The skipped modules are logged in `DEBUG`. I also added the modules that are skipped with the default settings to the issue and made it `breaking`.

Depends on https://github.com/interactive-instruments/xtraplatform/pull/158.
Depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/266.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1174.surge.sh

cportele (approved):

PR #1178, opened 2024-03-11
add popup for custom styles

### LINKED ISSUE #1153: Features HTML: add popup with the feature ID on custom styles
If a default style is used in the HTML representation of features, a popup with the feature ID is shown when the mouse pointer is over a feature geometry.

If a custom style is used, no such popup is shown, the [popup is only created, if no style URL is set](https://github.com/interactive-instruments/ldproxy/blob/master/ogcapi-stable/ogcapi-html/src/main/javascript/src/components/MapLibre/Configuration/index.jsx#L174-L190).

If it is a small effort, we should provide the same popup behavior also for custom styles.

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #1153.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1178.surge.sh

cportele (approved):
Looks good

PR #1181, opened 2024-03-21
null values

### LINKED ISSUE #1180: improve handling of `null` values
Currently, all `null` values emitted by a provider are removed in the transformation pipeline and will never reach the encoder. Furthermore, most encoders would ignore `null` values anyways.
This will introduce the following improvements:

- When a property is marked as [required](https://docs.ldproxy.net/providers/details/constraints.html), `null` values are always passed through to the encoder.
- All other `null` values are still removed by default, but that can be changed using the new `removeNullValues` transformation.
- The GeoJSON and JSON-FG encoders will add `null` values to the output instead of ignoring them.

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #1180.

See https://interactive-instruments-ldproxy-pr-1181.surge.sh/providers/details/transformations.html.

Depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/267.

### DIFF
--- a/ogcapi-stable/ogcapi-features-geojson/src/main/java/de/ii/ogcapi/features/geojson/app/GeoJsonWriterProperties.java
+++ b/ogcapi-stable/ogcapi-features-geojson/src/main/java/de/ii/ogcapi/features/geojson/app/GeoJsonWriterProperties.java
@@ -168,16 +168,20 @@ public class GeoJsonWriterProperties implements GeoJsonWriter {
   }
 
   protected boolean shouldSkipProperty(EncodingAwareContextGeoJson context) {
-    return !hasMappingAndValue(context) || (context.schema().get().isId() || context.inGeometry());
+    return !hasMapping(context) || (context.schema().get().isId() || context.inGeometry());
   }
 
-  protected boolean hasMappingAndValue(EncodingAwareContextGeoJson context) {
-    return context.schema().filter(FeatureSchema::isValue).isPresent()
-        && Objects.nonNull(context.value());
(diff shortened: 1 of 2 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1181.surge.sh

cportele:
@azahnen - I tried the following, but it does not seem to have an effect on the output, there are no `null` values. Did I misinterpret the documentation?

```yaml
  CulturePnt:
    sourcePath: /culturepnt
    type: OBJECT
    transformations:
      '*':
      - removeNullValues: false
    properties:
      ...
```

cportele:
It works as expected, if I add the transformation in the `GEO_JSON` or `JSON_FG` building blocks in services.

azahnen:
@cportele In the provider, the syntax should look like this:

```yml
CulturePnt:
    sourcePath: /culturepnt
    type: OBJECT
    transformations:
    - removeNullValues: false
    properties:
```

cportele:
Ah ok, I misread the text in the documentation.

cportele (approved):
Works as expected (once I figured out how to configure the type properly 😉). I also checked that there is no difference in CSV or FlatGeobuf.

I probably would have use something like `includeNullValues` with a default value `false`, but the result is the same.

azahnen:
@cportele Since the transformation pipeline is actively removing these properties/values, it seemed more fitting to describe this action in the transformation name. If we would instead have made this an option in the format building blocks, I would also have preferred `includeNullValues`.

PR #1191, opened 2024-03-28
change services endpoint path to root

### LINKED ISSUE #1190: change API catalog path from `/rest/services` to `/`
The API catalog should be available at the root path, and with the removal of the manager there are no longer any hindrances. The default URL will change from `http://localhost:7080/rest/services` to `http://localhost:7080`.

That means if ldproxy is operating behind a reverse proxy, the configuration of the reverse proxy has to be adjusted.
If for some reason `externalUrl` is set to the old default URL, the setting in `cfg.yml` has to be adjusted.

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #1190 .

Depends on interactive-instruments/xtraplatform#160.

Traefik can be used to test that using a reverse proxy to change the path still works as expected:

```yml
version: "3.3"

services:

  traefik:
    image: "traefik:v2.11"
    command:
      - "--providers.docker=true"
      - "--entrypoints.web.address=:7080"
    ports:
      - "7080:7080"
    volumes:
      - "/var/run/docker.sock:/var/run/docker.sock:ro"

  ldproxy:
    image: "docker.ci.interactive-instruments.de/iide/ldproxy:root-path"
    environment:
      - EXTERNAL_URL="http://localhost:7080/foo/bar" 
      - DB_HOST=host.docker.internal:5432
    volumes:
      - /path/to/configs-ldproxy/xleit:/ldproxy/data
    labels:
      - traefik.http.services.ldp.loadbalancer.server.port=7080
      - traefik.http.routers.ldp.rule=Host(`localhost`) && PathPrefix(`/foo/bar`)
      - traefik.http.routers.ldp.entrypoints=web
      - traefik.http.middlewares.prefix.stripprefix.prefixes=/foo/bar
      - traefik.http.routers.ldp.middlewares=prefix
```

### DIFF
--- a/demo/vineyards/README.md
+++ b/demo/vineyards/README.md
@@ -142,7 +142,7 @@ INFO  [2021-12-18 07:29:39,089]                vineyards - Feature provider with
 INFO  [2021-12-18 07:29:39,400]                vineyards - Service with id 'vineyards' started successfully.  
 ```
 
-Congratulations, you are now ready to use the API, just open the [landing page](http://localhost:7080/rest/services/vineyards) or the [vineyards features](http://localhost:7080/rest/services/vineyards/collections/vineyards/items).
+Congratulations, you are now ready to use the API, just open the [landing page](http://localhost:7080/vineyards) or the [vineyards features](http://localhost:7080/vineyards/collections/vineyards/items).
 
 Have a look at the updated configuration files. They should like the following, first the providers file, the the services file.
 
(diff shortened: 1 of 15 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1191.surge.sh

cportele (approved):
LGTM

More as a reminder, but we still have to update https://github.com/interactive-instruments/ldproxy/blob/root-path/src/markdown/docs/en/10-installation.md, which still references the manager as a test that a deployment has started.

PR #1203, opened 2024-04-16
Fix text search

cportele:
### Pull request checklist

-   [ ] Added/updated unit tests
-   [x] Updated documentation
-   [x] All checks are passing

### Changes introduced by this PR

Closes #1202

Documentation: An outdated limitation is also removed, the comparison is case-insensitive.

PR #1204, opened 2024-04-16
Fix text search, part 2

cportele:
### Pull request checklist

-   [x] Updated documentation

Follow-up to #1203. In that PR, the reference to the case limitation was only removed in the API definition, not in the ldproxy documentation.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1204.surge.sh

PR #1213, opened 2024-05-01
intersects parameter: quote property name to avoid keyword clashes

### LINKED ISSUE #1212: parameter 'intersects': allow primary geometry property names that are CQL2 keywords
If the primary geometry property has a name that is a CQL2 keyword (e.g. "bbox"), the property needs to be enclosed in double quotes in CQL2 Text. This is currently not the case in the filter generated from the `intersects` query parameter.

In addition, the test for a `NULL` value should be omitted, if the geometry property is required. See #1157.

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1212.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1213.surge.sh

PR #1224, opened 2024-06-12
add bounds to the vector source of a MapLibre style

### LINKED ISSUE #1223: MapLibre styles: add bounds property from the data
Adding a `bounds` property in a MapLibre style source can help clients to avoid sending requests for tiles outside of the spatial extent. Adding a bounding box manually can result in values that are out-of-sync with the spatial extent of the dataset or collection. 

Add an option to STYLES to add the `bounds` property on-the-fly to the vector tile source of a style when requested via the API, e.g., `addBoundsToVectorSource: true`.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1223.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1224.surge.sh

azahnen:
@cportele This looks good to me, but is there a reason to not always enable this? I don't see why this would be a breaking change, it shouldn't change functionality but improve the performance for most users. So I would enable it by default.

cportele:
@azahnen - Good question. Originally it was not clear to me whether it would be a good idea to always enable this or what side-effects this could result in. Therefore, I disabled it by default, but in testing it always improved behavior, so I agree, it makes sense to enable it by default. I will make the change...

PR #1235, opened 2024-07-17
check integer feature ids

### LINKED ISSUE #1234: Features: always return 404 for features that do not exist
In case the feature type has an id that is an integer and the requested featureId is not an integer, the result is an SQL error (because of the type mismatch). This has two issues:

- Status 400 is returned instead of a 404.
- Executing the SQL query should be avoided, this issue should be detected before.

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1234


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1235.surge.sh

azahnen (approved):
Works as expected.

PR #1239, opened 2024-08-02
add support for codelists in the API

### LINKED ISSUE #1238: Add support for codelists in the API
Codelists are currently internal resources and not visible to API users. Users can typically only implicitly observe codelists when they compare the values of feature properties in GeoJSON etc. vs the value in HTML. It would be valuable to make codelists explicit and support them in the API.

- Publish codelists as resources in the API in a new building block CODELISTS
  - `/codelists` provides the list of codelists in an API
  - `/codelists/{codelistId}` provides a codelist
  - All resources should be available in HTML and JSON
  - The codelists of an API are the codelists that are identified in schema constraints in the feature provider schema
- Add a `x-ldproxy-codelistUri` annotation in the schemas for properties with a codelist constraint, if CODELISTS is enabled
- Support new profiles "val-as-code" and "val-as-title" for the Features/Feature requests, if CODELISTS is enabled
  - Only for collections that use one or more codelists
  - "val-as-code" is the current behavior, if not codelist transformation is used 
  - "val-as-title" is the current behavior, if the codelist transformation is used
  - GML: only support "val-as-code" ("val-as-title" would break schema validity)
  - HTML: only support "val-as-title" (since human readability is essential for HTML); as a result, it is no longer necessary to configure codelist transformations in the FEATURES_HTML building block

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1238

Depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/278


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1239.surge.sh

azahnen:
@cportele
This did not work very well with some use cases yet, I pushed some changes:

- I fixed an NPE, it occurred presumably when requesting features for a type that does not have any profile-related properties at all.
- I added the missing JSON link on the codelist HTML page. To add the links in the JSON representation would indeed be a bit cumbersome, so I left that as it is for now.
- The endpoints now also consider codelists that are only referenced in provider transformations, but not in constraints. Format specific transformations are not yet considered, but that might make sense as well.
- Values do not have a unique id like entities, they have a path. That means at least for codelists that these can be grouped in subdirectories, e.g. `aaa/inspire/codelist1` is a valid codelist reference. The endpoints now also consider these nested codelists. I chose the simple approach of encoding the paths, so the id for the given example would be `aaa%2Finspire%2Fcodelist1`. A more advanced approach would represent the paths in the URL design, but then directory listings might be expected by users.

Some other thoughts:

- I do not like the format specific profile constraints. If somebody wants to see the codes in HTML, why not? And the schema validity argument for GML is not consistent with other parameters, 
projections currently do not respect that as well. So I would allow both profiles for all formats, only the default should be format specific.
- Speaking of defaults, these are not configurable yet, are they? This would also include the `ref-as-` or other profiles. That might be future work, but I think it would be good to have a `profiles` option for each format building block with a list of the default profiles, which could then be changed.
- It might be confusing for users when to use transformations or constraints or both. I guess the recommended way in the future would be constraints + profiles, so it might be good to explain that in the transformations and constraints docs.

cportele:
@azahnen 

I have updated the branch:

- Codelists referenced from the service configuration are now published, too.
- I have removed the default values for the "rel" and "val" profiles and replaced them with format-specific default values.
- The documentation has been updated:
  - Recommendation to specify codelists as constraints, not transformations. Only the constraints can be controlled by the profile.
  - The documentation of the profile parameter has been updated to reflect the changes, too.

One comment on your edits:

- The use of `PercentCodec.decode()` seems superfluous, the values are automatically decoded. I have not removed it, because it was not clear to me, if I am missing something.

azahnen (approved):
This looks good to me now. I pushed some more improvements:

- I removed the PercentCodec usage, it seems to work fine without explicit encoding and decoding.
- I decoupled the profiles and formats in the profile negotiation. I first defined the supported and default profile values in the `FeatureFormat` classes, but that just replaced the hardcoded media types in the `Profile` classes with hardcoded profile values in the `FeatureFormat` classes. Therefore I introduced an approach where the `Profile` classes determine the supported and default values based on characteristics provided by the `FeatureFormat` classes, currently `isComplex` and `isForHumans`.
- I also made the default profile values configurable, this is available for all format specific building blocks. For example you could change the GeoJSON default profiles like this:
  ```yml
  - buildingBlock: GEO_JSON
    defaultProfiles:
      val: val-as-title
      rel: rel-as-uri
  ```

PR #1244, opened 2024-08-03
update to align with Features Part 6

### LINKED ISSUE #1240: Projections: upgrade to current draft of Features Part 6
Upgrade the PROJECTIONS building block to the current draft of [OGC API Features Part 6: Property Selection](https://docs.ogc.org/DRAFTS/24-019.html):

- Deprecate `skipGeometry`
- Include the geometry properties in `properties`
- Add `exclude-properties` parameter (mutually exclusive with `properties`)
- Add conformance class URIs

cportele:
### Pull request checklist

-   [x] Updated documentation

deprecate support for `skipGeometry`, include spatial properties in `properties`, add support for `exclude-properties`, add conformance class URIs

closes #1240

depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/279


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1244.surge.sh

azahnen (changes requested):
This works mostly as expected, but I spotted some potential issues:

- `exclude-properties` will always exclude the geometry, even if it is not part of the list. I guess that is not intentional?
- The OpenAPI schema for the parameter was not adjusted to contain the geometry.
- This is not a new issue, but the parameter values are not validated. That might not have to be implemented right now, but before we move this to mature/stable.

cportele:
@azahnen 

The issues should be fixed.

The parameter values were validated, but the schema validation did not work properly for query parameters with array values, most values were accepted.

azahnen (approved):

PR #1253, opened 2024-08-07
improve "result-type" parameter

### LINKED ISSUE #706: Support for hitsOnly
### ldproxy Version

master

### Current Behavior

Originally the API supported a request to not return the features, but just the number of features that would be returned. The `hitsOnly` option is still in `FeatureQuery` and the `TransformationContext` classes, but support for it was removed at some point in the overhaul of the transformation pipeline implementation. Such a capability is currently not supported by OGC API Features, but may still be helpful. The capability should at least be restored, even if it is not available via the API at this point. It can also be useful internally.

There are also the options `propertyOnly` and `onlyHitsIfMore` (the latter only in `TransformationContext`), which also have no effect. If these are no longer needed, they should be removed.

### Expected Behavior

`hitsOnly` is supported again.

### Steps to Reproduce

Execute a query internally with `hitsOnly` set to `true`. The setting has no effect.

### Relevant log output

```shell
n/a
```

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #706

Depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/281

One comment for discussion: "result-type" is not specified in the OGC API specs, but it has always been in FEATURES_CORE. Should this parameter be moved to its own module in ogcapi-custom?

### DIFF
--- a/.woodpecker/release.yml
+++ b/.woodpecker/release.yml
@@ -47,6 +47,8 @@ steps:
       platforms:
         - linux/amd64
         - linux/arm64
+## TODO: latest not updated
+## workaround: docker buildx imagetools create --tag "iide/ldproxy:latest" "iide/ldproxy:4.1.0"
 
   github:
     image: node:18
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Changelog
 
+## v4.1.0 (05/08/2024)
+#### Implemented enhancements
+
+-  Tiles: support raster tile generation with xtratiler [#1251](https://github.com/interactive-instruments/ldproxy/issues/1251)
+-  add Oracle Feature Provider [#1232](https://github.com/interactive-instruments/ldproxy/issues/1232)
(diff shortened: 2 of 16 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1253.surge.sh

azahnen (approved):
LGTM. I guess we put it into FEATURES_CORE since we expected it to be included in the spec someday. If that will not happen in the foreseeable future, I think we should move it to ogcapi-custom.

cportele:
@azahnen - Since there is no active discussion at the moment, I have moved the query parameter to a new RESULT_TYPE building block in ogcapi-custom.

cportele:
@azahnen - the CI build fails with a vuepress error, but I do not see what the issue is and locally I do not get an error. Do you have an idea, what goes wrong?

azahnen:
@cportele The issue was in the `labelTemplate` docs. For some reason pipes have to be escaped like this `\|`, an example is also in the `stringFormat` docs. The build was successful now.

cportele:
@azahnen - Thanks!

PR #1263, opened 2024-08-26
maintain CRS dimension in cached spatial extents

cportele:
### Pull request checklist

-   [ ] Updated documentation

If a changing spatial extent is updated, it is stored as a CRS84 extent. If the data is 3D, the extent must be a CRS84h extent, otherwise the next update will fail, because the delta extent is in CRS84h, but the cached extent is in CRS84.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1263.surge.sh

PR #1270, opened 2024-09-05
Root concat

### LINKED ISSUE #1269: support concat on root level
Currently `concat` is not allowed on the root level. This will add support for that, so that feature instances for a single  type may originate from more than one table.

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #1269.

Depends on interactive-instruments/xtraplatform-spatial#285

**Todo**

- [x] docs

**Example**

```yml
administrativeunit:
  type: OBJECT
  concat:
  - sourcePath: "/o73005{sortKey=sch}"
    type: OBJECT
    properties:
      id:
        sourcePath: sch
        type: STRING
        role: ID
  - sourcePath: "/o73004{sortKey=sch}"
    type: OBJECT
    properties:
      id:
        sourcePath: sch
        type: STRING
        role: ID
```

### DIFF
--- a/ogcapi-custom/ogcapi-features-csv/src/main/java/de/ii/ogcapi/features/csv/app/FeaturesFormatCsv.java
+++ b/ogcapi-custom/ogcapi-features-csv/src/main/java/de/ii/ogcapi/features/csv/app/FeaturesFormatCsv.java
@@ -61,13 +61,11 @@ public class FeaturesFormatCsv extends FeatureFormatExtension implements Conform
           .parameter("json")
           .build();
 
-  private final FeaturesCoreProviders providers;
   private final FeatureSchemaCache schemaCache;
 
   @Inject
   public FeaturesFormatCsv(FeaturesCoreProviders providers, ExtensionRegistry extensionRegistry) {
-    super(extensionRegistry);
-    this.providers = providers;
+    super(extensionRegistry, providers);
     this.schemaCache = new SchemaCacheSfFlat();
   }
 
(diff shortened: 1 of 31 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1270.surge.sh

cportele:
I created a new type `Aeronautic in the daraa configuration as a concat of `AeronauticCrv` and `AeronauticSrf`. I then accessed various resources of the API with the following observations:

- Accessing feature resources was successful in formats HTML, GeoJSON and JSON-FG.
- In FlatGeobuf (and other formats that require a fixed schema), features that do not fully fit in the schema of the first option result in errors. My guess is that these encodings must be disabled for collections that use a root concat type.
- The schema resource of the collection only represents the first option.
- The list of queryables for the collection was empty although three queryables were defined in the service configuration. 
- Using such a type in a tile provider failed with the following error when accessing a tile: "Filter is invalid. Unknown property: 0_geometry."
- The schema in the tileset metadata only represents the first option.

azahnen:
@cportele 
The issues should be fixed. 
For feature formats I introduced `supportsRootConcat`, which is currently only set to `true` for JSON-based formats and HTML. It would most likely be enough to use the union of all properties for the other formats, but for now they are disabled with a warning.

cportele:
@azahnen 
There is still at least one issue with tiles. The tileset metadata states that all geometries are lines. This is incorrect as some features have polygons. This has the side-effect that the default wireframe style only shows the line string features, the polygon features are not rendered.

The cause is that geometry type / dimension is determined from the first primary geometry property. This was correct when there could only be one, but with concat there can now be multiple.

It is easy to fix in this case, but we probably need to review all cases where a similar schema constraint is assumed, which no longer applies. Should I work on this?

azahnen:
@cportele 
Maybe it would make sense to adjust the `MappingOperationResolver` so that the types of the primary geometries are already aligned when used by any downstream cases? 
If all primary geometries have the same type, nothing would be changed, otherwise all geometries would get the type `ANY`.

azahnen:
@cportele 
True. I think then it is best if you work on this.

cportele:
@azahnen 
OK, will do.

cportele (approved):

PR #1275, opened 2024-10-05
fix query parameter applicability

cportele:
### Pull request checklist

-   [x] Updated documentation

The methods that determine whether a query parameters is enabled / applicable for an API / collection were not always correct and use consistently. For example, the `intersects` parameter was enabled when the FEATURE_EXTENSIONS module was enabled, but the `intersectsParameter` option was not enabled. One reason was too much duplicated code. 

This commit uses defaults that apply to most cases so that only special cases need to override the default methods. The only aspect that varies for each parameter are the paths to which the query parameter applies, this is now declared in a separate method.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1275.surge.sh

azahnen (approved):
LGTM

Issue #1284, opened 2024-10-17
Improve performance when requesting many features


cportele:
Currently the SQL schema is processed for every row in a features request, but the processing could be done once before the SQL result rows are processed. This should be changed. 

This behavior was introduced in v3.6.

PR #1287, opened 2024-10-26
Improve vector tile seeding performance for sparse datasets

cportele:
### Pull request checklist

-   [x] Updated documentation

This pull request improves the vector tile seeding performance for some datasets. It includes two optimizations:

1. Avoid unnecessary deep object copies: When preparing the SQL query information, the relevant `SchemaSql` instances were copied (a deep copy). This overhead is in particular costly for a large number of queries with smaller result sets. A typical example is vector tile seeding in larger scales. This change reuses the existing immutables instead of copying them.

2. Avoid unnecessary feature queries for vector tiles: For tilesets that are marked as sparse in the tile provider (via a new option `sparse`), no feature query is executed and processed, if the feature type has no features or if the tile does not intersect the spatial extent of the feature type. An empty tile is immediately returned instead. 
 Note that this could be optimized further, if the spatial extent would not be a single bounding box, but could express multiple areas with data as it is also foreseen in OGC API Features.

The changes improved the seeding time in a test case from more than 3 hours to 45 minutes (roughly 1 million tiles; 234 tilesets; 1 combined tileset; only a third of the feature types have features and others are only sparsely populated).

Depends on https://github.com/interactive-instruments/xtraplatform-spatial/pull/290

### DIFF
--- a/ogcapi-custom/ogcapi-codelists/src/main/java/de/ii/ogcapi/codelists/app/CodelistFormatHtml.java
+++ b/ogcapi-custom/ogcapi-codelists/src/main/java/de/ii/ogcapi/codelists/app/CodelistFormatHtml.java
@@ -18,6 +18,7 @@ import de.ii.ogcapi.foundation.domain.I18n;
 import de.ii.ogcapi.foundation.domain.Link;
 import de.ii.ogcapi.foundation.domain.OgcApiDataV2;
 import de.ii.ogcapi.foundation.domain.URICustomizer;
+import de.ii.ogcapi.html.domain.FormatHtml;
 import de.ii.ogcapi.html.domain.HtmlConfiguration;
 import de.ii.ogcapi.html.domain.NavigationDTO;
 import de.ii.xtraplatform.codelists.domain.Codelist;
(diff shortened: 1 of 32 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1287.surge.sh

cportele:
I also did test runs seeding vector tiles for the demo datasets (daraa, strassen, vineyards), up to zoom level 12, using docker on macOS. The Daraa tilesets were set to sparse. 

The improvements over version 4.0/4.1 are likely the result of #1284. 

The improvements in the combined Daraa tiles are likely the result of the optimization for sparse data in this PR as until now many empty tilesoutside of the tileset area are generated when combining tiles. The overall effect of the schema copy optimization is relatively small as tiles were only seeded to level 12 in the tests.

  | 4.0 | 4.1 | master | perf2
-- | -- | -- | -- | --
Collections (Vineyards, Strassen, Daraa) | - | 27:08 | 19:38 | 19:05
Daraa Combined | - | 1:55 | 1:52 | 0:49
Strassen Combined | - | 0:50 | 0:51 | 0:47
Total | 28:35 | 27:58 | 20:29 | 19:53

PR #1292, opened 2024-11-06
fix temporal extent on landing page

cportele:
### Pull request checklist

-   [ ] Updated documentation

There were two issues with the temporal extent on the landing page:

- unbounded interval ends were not handled properly;
- the JSON representation did not write ISO 8601, but milliseconds since epoch.

Since including this information in the landing page is an ldproxy extension (so that it can be shown in the HTML representation), this issue has not been noticed yet.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1292.surge.sh

PR #1295, opened 2024-11-12
render immutable view

### LINKED ISSUE #1294: Features HTML: mapPosition 'AUTO' is slow
When a feature page has many features (e.g. 1000 features), rendering the page is very slow, if `mapPosition: AUTO` (the default) is used. This is caused by analysing all features for every execution of  the "derived" values for `isMapRight()` or `isMapTop()`.  Since the view is a Modifiable not an Immutable, the value is not just computed once, but for every invocation.

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1294. Before rendering, derive an immutable view from the modifiable view to leverage the "derive" annotations in the evaluation.


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1295.surge.sh

azahnen (approved):
LGTM (You could just use `renderView(transformationContext.collectionView().toImmutable())`, but this works the same.)

PR #1297, opened 2024-11-13
Issue 1262

### LINKED ISSUE #1262: Filter Editor: fields are not in alphabetic order
### ldproxy Version

3.6.1.

### Current Behavior

_I also posted this in the discussions section (#1255), but the difference between how this works when I specify the fields vs. when I use "include all", makes me believe that this isn't intended behavior. Something similar was also perceived as a bug/regression in #795._

I'm having some problems getting the filter fields in alphabetic order. That is, I'm having problems when I'm using "include all" like this:

    buildingBlock: QUERYABLES
    enabled: true
    included:
        "*"

If I specify what fields I want to include, e.g.:

included:
- vernetype
- navn

the filter fields are sorted.

### Expected Behavior

I expect that the filter fields are rendered in alphabetical order, regardless of if I specify the fields I want to be included or if I specify to include all existing fields.

### Steps to Reproduce

Adding the following in a config-file under defaults or for a specific service.
```
 - buildingBlock: QUERYABLES
   enabled: true
   included:
     - "*"
```

### Relevant log output

_No response_

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #1262.

### DIFF
--- a/ogcapi-stable/ogcapi-html/src/main/javascript/src/components/FilterEditor/Editor/Fields/index.jsx
+++ b/ogcapi-stable/ogcapi-html/src/main/javascript/src/components/FilterEditor/Editor/Fields/index.jsx
@@ -27,7 +27,7 @@ const FieldFilter = ({
 
   const filtersToMap = Object.keys(filters).filter(
     (key) => filters[key].remove === false && key !== "bbox" && key !== "datetime"
-  );
+  ).toSorted();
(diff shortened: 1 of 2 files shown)


azahnen:
Deploying preview to https://interactive-instruments-ldproxy-pr-1297.surge.sh

PR #1309, opened 2024-12-08
Build plugins

azahnen:
I completely overhauled the gradle build plugins, here is a list of the main changes:

- Java language level updated to 17
  - JDK 21 may be used
  - language level will be updated to 21 soon, just have to make sure that `ldproxy-cfg` stays on 17
- improved embedding of third party libraries
  - there should be no `tpl` errors anymore that require a `clean`
- improved caching of many tasks
  - in many cases the builds should be faster
- improved IntelliJ integration
  - jars with third party libraries are built when loading the project, so the source code shouldn't show any errors
  - automatically create `run` and `debug-external` run configurations

Depends on interactive-instruments/xtraplatform#197
Depends on interactive-instruments/xtraplatform-spatial#314


azahnen:
@cportele 
Can you check if this works for you?
You should clean all projects before updating and maybe also delete the `.idea` folders if you don't have any sophisticated run configurations (Or just checkout new copies).

azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1309.surge.sh

cportele (approved):
@azahnen - Builds work without issues for me

PR #1312, opened 2024-12-17
support temporal queryables as parameters

### LINKED ISSUE #1308: query parameters for DATE/DATETIME queryables always result in an empty set
Query parameters for queryables are only supported for the value types STRING, INTEGER, FLOAT, BOOLEAN. For queryables of DATE or DATETIME, the query parameter is created, but always evaluates to `false`. 

Either DATE/DATETIME queryables should be supported or no query parameter should be created for those queryables.

cportele:
### Pull request checklist

-   [ ] Updated documentation

Map query parameters with DATE or DATETIME values to CQL2.

Closes #1308


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1312.surge.sh

PR #1333, opened 2025-02-06
remove styles http access from backend

### LINKED ISSUE #1332: style errors with http client proxy
### ldproxy Version

4.2.1

### Current Behavior

When ldproxy only has network access through a http proxy, the map client on the items page will never use a configured style. It prints in the log that it cannot be accessed and falls back to default wireframe styles.

### Expected Behavior

Map clients should use correctly configured styles when running behind a http proxy.

### Steps to Reproduce

Setup an instance that only has network access through a http proxy, configure an HTML default style, open the items page.

### Relevant log output

```shell

```

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Fixes #1332.

### DIFF
--- a/ogcapi-draft/ogcapi-tiles3d/src/main/java/de/ii/ogcapi/tiles3d/app/Format3DTilesTilesetHtml.java
+++ b/ogcapi-draft/ogcapi-tiles3d/src/main/java/de/ii/ogcapi/tiles3d/app/Format3DTilesTilesetHtml.java
@@ -19,8 +19,8 @@ import de.ii.ogcapi.foundation.domain.OgcApi;
 import de.ii.ogcapi.foundation.domain.OgcApiDataV2;
 import de.ii.ogcapi.foundation.domain.URICustomizer;
 import de.ii.ogcapi.html.domain.HtmlConfiguration;
-import de.ii.ogcapi.styles.domain.StyleFormatExtension;
-import de.ii.ogcapi.styles.domain.StyleRepository;
+import de.ii.ogcapi.html.domain.MapClient;
+import de.ii.ogcapi.html.domain.StyleReader;
 import de.ii.ogcapi.tiles3d.domain.Format3dTilesTileset;
 import de.ii.ogcapi.tiles3d.domain.Tiles3dConfiguration;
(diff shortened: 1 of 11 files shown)


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1333.surge.sh

azahnen:
@cportele 
There were two places where styles were accessed via http from the backend. 
The first one just checked for its existence. This one did not use the correct html client and caused the error. I replaced that with a check on the value store.
The second one injected 3dtiles-styles into HTML pages for cesium. I replaced that with a fetch on the client side, like it is handled for MapLibre as well.
By the way, both requests also did not work when enabling access control for the items page.

Can you check if everything still works as expected for you?

cportele (approved):
The update of the 3D Tiles tileset view was missing, I have updated the code. Everything else worked as expected.

cportele:
@azahnen 

This change has introduced a new issue that I did not discover in my review. 

For collections, the styles are in general derived from the dataset style by removing all layers for other collections. In the StyleReader, the derived collection styles do not exist, so the style always falls back to the dataset style. For GeoJSON data, each feature is now rendered using the layers for all collections, which results in incorrect maps.

The problem is that the information whether collection styles are derived from the dataset style is only available in the STYLES module. That is, the FEATURES_HTML module as well as `StyleReaderImpl` has no information, if a derived collection style is available.

Probably the best option would be 

* to deprecate `deriveCollectionStyles`;
* derivation depends on the style format only (always `true` for MapLibre styles and otherwise `false`);
* update `StyleReader` to include the derivation logic.

This would likely require a 4.2.3 release as most APIs that have multiple collections and use MapLibre styles for feature data will be affected.

The alternative would be to revoke the changes in this PR.

Any thoughts?

azahnen:
@cportele 
Both approaches sound reasonable, we should discuss that in more detail. I think we should implement that vor 4.3.0.

As a quick hack for 4.2.3, we could serialize `apiData` and then deserialize it again as generic map to get access to `deriveCollectionStyles`.

cportele:
@azahnen - That would work. Let's discuss that tomorrow.

PR #1349, opened 2025-02-23
Improve support for ad-hoc compound CRS

### LINKED ISSUE #1348: Improve support for ad-hoc compound CRS
A Compound CRS as a combination of a 2D horizontal CRS and a 1D vertical CRS is supported as the native CRS in a feature provider (via `verticalCode`). 

Other API elements, however, that can express such CRSs are not yet supported:
- The query expression in the SEARCH module only supports a single CRS (`crs`). An optional `verticalCrs` member should be added for cases where `crs` is a 2D horiziontal CRS. 
- The JSON-FG `coordRefSys` member can express an ad-hoc compound CRSs, but `verticalCode` is not yet considered when serializing the `corrdRefSys` member. This should be corrected.

In other cases, standard API building blocks do not support such CRSs where the CRS is referenced by an identifier. Where this is a single identifier, we cannot support ad-hoc compound CRSs as this would break conformance to the relevant standard. These cases are:

- Query parameters `crs`, `bbox-crs` and `filter-crs`;
- the CRS lists in the Feature Collections and Feature Collection resources;
- the `Content-Crs` header;
- the CRS of a TileMatrixSet.

In addition, some formats do not support ad-hoc compound CRSs where the CRS is referenced by a single identifier. These formats are: FlatGeoBuf, CityJSON, GML.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1348.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/328.

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '6.3.0-SNAPSHOT'
 xtraplatform-native  = '2.5.0-SNAPSHOT'
-xtraplatform-spatial = '7.3.0-issue-1348-SNAPSHOT'
+xtraplatform-spatial = '7.3.0-SNAPSHOT'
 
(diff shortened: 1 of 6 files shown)


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1349.surge.sh

azahnen (dismissed):
LGTM

PR #1355, opened 2025-03-04
Generally support parameters in stored queries - not only in filters

cportele:
### Pull request checklist

-   [x] Updated documentation

Currently, parameters are only supported in filter expressions in a stored query. Parameters are also useful in other members of a query expression, e.g. the CRS information, the limit value or the profiles to apply.

This commit supports the use of parameters in all members of the query expression of a stored query, except "id", "title", "description", and "offset".

A new immutable `StoredQueryExpression` is introduced that supports parameter objects in the values of the query expression of a stored query. These instances are what is stored as values. The `QueryExpression` is simplified to remove all support for parameters.

Example of the use of parameters in "crs" and "profiles":

```
    ...,
    "crs": {
        "$parameter": {
            "crs": {
                "title": "Koordinatenreferenzsystem",
                "description": "Koordinatenreferenzsystem in dem Geometrien zurückgegeben werden sollen.",
                "type": "string",
                "enum": [
                    "http://www.opengis.net/def/crs/EPSG/0/25832" ,
                    "http://www.opengis.net/def/crs/EPSG/0/4258"  ,
                    "http://www.opengis.net/def/crs/OGC/1.3/CRS84"
                ],
                "default": "rhttp://[www.opengis.net/def/crs/EPSG/0/25832](http://www.opengis.net/def/crs/EPSG/0/25832)"
            }
        }
    },
    "profiles": [
        {
            "$parameter": {
                "profil-rel": {
                    "title": "Anzuwendendes Profil für Relationen",
                    "description": "Das Profil für die Darstellung von Objektreferenzen, das angewendet werden soll. `rel-as-key` bedeutet, dass die Objektreferenzen als Fremdschlüssel des referenzierten Objekts dargestellt werden; bei `rel-as-uri` wird die HTTP-URI des referenzierten Objekts verwendet; bei `rel-as-link` wird ein Link-Objekt verwendet.",
                    "type": "string",
                    "enum": ["rel-as-key", "rel-as-uri", "rel-as-link"],
                    "default": "rel-as-uri"
                }
            }
        },
        {
            "$parameter": {
                "profil-val": {
                    "title": "Anzuwendendes Profil für Codelisten",
                    "description": "Das Profil für die Darstellung von Werten aus Codelisten, das angewendet werden soll. Bei `val-as-code` werden die Werte als  dargestellt; bei `val-as-title` werden die Werte als lesbare Texte dargestellt.",
                    "type": "string",
                    "enum": ["val-as-code", "val-as-title"],
                    "default": "val-as-code"
                }
            }
        }
    ],
    ...
```

### DIFF
--- a/ldproxy-cfg/src/main/java/de/ii/ldproxy/cfg/LdproxyCfgImpl.java
+++ b/ldproxy-cfg/src/main/java/de/ii/ldproxy/cfg/LdproxyCfgImpl.java
@@ -16,6 +16,7 @@ import com.networknt.schema.JsonSchemaFactory;
 import com.networknt.schema.SpecVersion.VersionFlag;
 import com.networknt.schema.ValidationMessage;
 import de.ii.ogcapi.features.search.domain.QueryExpression;
+import de.ii.ogcapi.features.search.domain.StoredQueryExpression;
 import de.ii.ogcapi.foundation.domain.OgcApiDataV2;
 import de.ii.ogcapi.styles.domain.MbStyleStylesheet;
 import de.ii.xtraplatform.base.app.StoreImpl;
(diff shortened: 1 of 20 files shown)


azahnen (approved):
LGTM 
Before we promote the module to `CANDIDATE`, we should think about a cleaner (de)serialization solution, but I think it is fine for now.

PR #1359, opened 2025-03-20
Reduce number of glTF drawing calls

### LINKED ISSUE #1358: Improve 3D Tiles / glTF performance
Currently each (building) feature is represented as a glTF node. Rendering performance, in particular in tiles with many features, can be improved. For example, by reducing the number of nodes/meshes/primitives - and as a result the drawing calls required to render the meshes.

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1358.

These changes join nodes/meshes/primitives to significantly reduce the number of drawing calls.

Additional improvements could be implemented, e.g., reordering of triangles for improved GPU performance or transmission size, compression, simplification of triangles, etc.


cportele:
To test the improvement, I generated a glb file with all 10500 buildings with and without the changes. I then used both versions in the babylon.js sandbox. The model with the changes renders at the full 120 fps when rotating the model. The model without the changes renders at around 13 fps (and feels sluggish).

PR #1370, opened 2025-03-27
replace instead of update count and extents

### LINKED ISSUE #1369: feature count adds up on dataset changes
### ldproxy Version

4.3.0

### Current Behavior

When the collection statistics are refreshed after a dataset change event, for example using the new periodic sync introduced in #633, the new feature count is added to the old one, so it increases every time.

### Expected Behavior

The new feature count should replace the old feature count.

### Steps to Reproduce

Trigger a dataset change event and check the feature count in the html view.

### Relevant log output

```shell

```

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #1369.


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1370.surge.sh

cportele (approved):
Works now as expected. I think I did not notice this when reviewing the original PR, because the log always had the correct value.

PR #1372, opened 2025-04-03
Tile generation: only set or exclude properties, if explicitly requested

### LINKED ISSUE #1371: tile cache is ignored for collection tiles
### ldproxy Version

4.3.0 (most likely exists since 4.2.0, #1244)

### Current Behavior

When requesting a collection tile and `PROJECTIONS` is enabled, it is always generated on the fly instead of being served from the cache.

### Expected Behavior

It should be served from the cache as long as no query parameters are used.

### Steps to Reproduce

Compare these two requests:

```sh
curl -o all.mvt https://demo.ldproxy.net/vineyards/tiles/WebMercatorQuad/8/87/133?f=mvt
```

```sh
curl -o coll.mvt https://demo.ldproxy.net/vineyards/collections/vineyards/tiles/WebMercatorQuad/8/87/133?f=mvt
```

The first one is much faster since it is served from the cache.

The reason ist that the query parameter `exclude-properties` is allowed for collection tiles and even if not set it will add tile generation parameters to the internal query, which in turn disables cache usage.

To fix this, I think the following method has to be adjusted so that it does not call `userParametersBuilder::addFields` when the parameter is empty.

https://github.com/ldproxy/ldproxy/blob/e1d5c0920a58cb25f30c1e2c24ee8ee21683db53/ogcapi-draft/ogcapi-projections/src/main/java/de/ii/ogcapi/projections/app/QueryParameterExcludeProperties.java#L178-L192

### Relevant log output

```shell

```

cportele:
### Pull request checklist

-   [ ] Updated documentation

Thanks for the analysis, @azahnen.

Closes #1371

### DIFF
--- a/ogcapi-draft/ogcapi-projections/src/main/java/de/ii/ogcapi/projections/app/QueryParameterExcludeProperties.java
+++ b/ogcapi-draft/ogcapi-projections/src/main/java/de/ii/ogcapi/projections/app/QueryParameterExcludeProperties.java
@@ -179,16 +179,19 @@ public class QueryParameterExcludeProperties extends OgcApiQueryParameterBase
       ImmutableTileGenerationParametersTransient.Builder userParametersBuilder,
       QueryParameterSet parameters,
       Optional<TileGenerationSchema> generationSchema) {
-    generationSchema
-        .map(TileGenerationSchema::getProperties)
-        .map(Map::keySet)
+    parameters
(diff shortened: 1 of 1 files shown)


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1372.surge.sh

azahnen (approved):

PR #1374, opened 2025-04-15
fix bug: "label" has no effect

### LINKED ISSUE #1373: Layer control: "label" has no effect
### ldproxy Version

4.3.0

### Current Behavior

Even if a layer in a layer control element of a MapLibre style includes a "label" member, the "id" is used in the legend.

### Expected Behavior

If "label" is set for a layer, the label should be used, not the layer id.

### Steps to Reproduce

In a layer control add a different label for the layer and check the layer control/legend.

### Relevant log output

```shell

```

PZahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #1373.


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1374.surge.sh

cportele (changes requested):
The label is now used, but the changes introduce a new issue. 

If the same label is used multiple times (with different ids), enabling or disabling a layer now also enables or disables other layers with the same label. The label is only for the presentation, internally the id has to be used to identify a group/layer or whether a group/layer is opened or enabled.

cportele (dismissed):
Looks good now

PR #1376, opened 2025-04-16
support ALIKE() in queryables parameters

### LINKED ISSUE #1375: CQL2: Predicate to filter array values with LIKE

Currently, queryables that are of type VALUE_ARRAY can only be filtered using the array operations (A_CONTAINS, etc.). This supports evaluating discrete sets of values. However, it does not support filtering of string arrays using a LIKE pattern. 

### Describe the Use Case

For example, if a building can have multiple addresses and one wants to select all buildings in a street that starts with "A". This is not supported in standard CQL2.

The CQL2 mechanism for extensions are functions. Add a function `ALIKE(array_property, like_pattern)`, where `array_property` is a queryable of type VALUE_ARRAY and `like_pattern` is a LIKE pattern.

For example: `ALIKE(address.street, 'A%')`

`ALIKE()` should also be used in query parameters of queryables that are arrays, if the value includes a "*".

### Additional Information

_No response_

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1375

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/335


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1376.surge.sh

azahnen (approved):
Works as expected.

PR #1391, opened 2025-05-16
PostGIS ST_AsMVT

### LINKED ISSUE #1390: support vector tile generation with ST_AsMVT from PostGIS

The current implementation is necessary to support complex mappings with joins, nested structures or value transformations. 
But for simple cases, using the builtin PostGIS functions should be a lot faster.

### Describe the Use Case

Much better performance for simple mappings.

_No response_

### Additional Information

_No response_

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #1390.

Depends on interactive-instruments/xtraplatform-spatial#344.

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '6.4.0-SNAPSHOT'
 xtraplatform-native  = '2.5.0'
-xtraplatform-spatial = '7.4.0-pgis-asmvt-SNAPSHOT'
+xtraplatform-spatial = '7.4.0-SNAPSHOT'
 
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '6.4.0-SNAPSHOT'
 xtraplatform-native  = '2.5.0'
-xtraplatform-spatial = '7.4.0-SNAPSHOT'
+xtraplatform-spatial = '7.4.0-pgis-asmvt-SNAPSHOT'
 


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1391.surge.sh

azahnen:
@cportele
The optimized vector tiles generation was implemented as an extension that has to be enabled in the feature provider, see the [documentation](https://ldproxy-ldproxy-pr-1391.surge.sh/de/providers/feature/90-extensions/postgis_tiles.html).

Regarding the [benchmark from the master thesis](https://github.com/FabianRechsteiner/vector-tiles-benchmark?tab=readme-ov-file#results), I was not able to replicate similar results, but I think that is because my hardware is a lot faster. These are the results comparing Martin with ldproxy without the optimization:

| | Test 1 | Test 2 | Test 3 | Test 4 | Test 5 |
| - | - | - | - | - | - |
| martin cache   |  38.55 |  20.89 |   9.97 |  35.24 |   4.77 |
| ldproxy master | 555.32 | 271.39 | 330.08 | 242.06 | 706.36 |

What I found out is that this comparison is pretty much useless since Martin uses an [in-memory tile cache](https://github.com/maplibre/martin/blob/c833034e0c79bcd1e69ea50bda31133e9aee2b2c/docs/src/config-file.md?plain=1#L38-L39). The result times each show the average of 100 request, which means for Martin that the tile is only generated once and then served from the cache.

I repeated the tests with a disabled cache for Martin and also added the optimized ldproxy:

| | Test 1 | Test 2 | Test 3 | Test 4 | Test 5 |
| - | - | - | - | - | - |
| martin no cache | 274.64 | 143.37 | 263.19 |  76.32 |  51.57 |
| ldproxy master  | 555.32 | 271.39 | 330.08 | 242.06 | 706.36 |
| ldproxy asmvt   | 310.34 | 164.60 | 367.91 |  53.41 | 284.40 |

This looks more reasonable, the optimized ldproxy is at least in the ballpark of Martin, which should be sufficient.

What is much more important than this synthetic benchmark is the real-world performance. Tile seeding with 16 threads for a customers dataset was 4 times faster on my machine.

cportele:
@azahnen 

I have made tests with various configurations and committed a number of fixes and improvements. 

Fixes:

- fix handling of empty tiles that are outside of the collection extent
- use the tile envelope for ST_AsMVTGeom, not the clip envelope
- use the clip envelope for ST_INTERSECTS to avoid "outside of domain" exceptions
- apply tile generation parameters in seeding jobs (needed for the clip envelope)
- support uppercase in property names
- use the tile extent from tile matrix set
- use the same buffer as in the default tile generator
- set the feature id, if it is an integer, as in the default tile generator
- set the type name as the name of the tile
- support CRS84, the default native CRS (LON_LAT is the default axis order in PGIS, only LAT_LON should be blocked)

Improvements:

- exclude properties that are not returnables from the analysis
- in log messages about properties include the type name
- clarify that properties with transformations are not excluded, only the transformations are not executed in WARN/IGNORE mode
- remove dead code
- update documentation to state that profiles are ignored in optimized tile generation
- improve error message for SQL exceptions

Two observations:

- Tile sizes for line strings and polygon geometries is often higher (sometimes double the size), the default encoder seems to simplify the geometry more than `ST_AsMVTGeom()`.
- For dataset tilesets with sparse data and many layers (e.g., daraa) the improvements are minimal. Total time with 8 threads is 5m12s vs 5m37s - with 4m30s used to combine the collection tiles. For cases where one only wants a dataset tileset and no collection tiles, it would be much faster to directly create the dataset tiles using UNION queries.

I will do more tests.

cportele:
@azahnen 

I have added more fixes/improvements:

- The optional list of properties in the tile provider for each tileset layer is now also taken into account.
- The documentation of the tile generation options have been updated to indicate whether they are taken into account in the asMVT option.
- So far, the code assumed that all asMVT options use the same feature provider and that all APIs use the same tile builder. This has been updated so that the tile builder and the SQL connector can vary per feature provider.

cportele (dismissed):
This should now be ready to be merged. 

There as still a few new TODO markers for code quality improvements or optimizations. Should we take care of them before merging or leave them in the code for later?

PR #1401, opened 2025-06-05
fix collection id extraction

### LINKED ISSUE #1400: http 500 when using filter obligation
### ldproxy Version

4.3.2

### Current Behavior

When using access control policies that return a filter obligation, the items request fails with an http error 500. This seems to happen only when the collection id equals the api id.

The reason is a bug in `ApiRequestContext.getCollectionId()`. This might also cause issues in other contexts.

### Expected Behavior

The request with the filter obligation should work without an error.

### Steps to Reproduce

Integrate a PDP that returns a filter obligation, request items.

### Relevant log output

```shell

```

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #1400.


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1401.surge.sh

PR #1409, opened 2025-06-20
Crud advanced

### LINKED ISSUE #1408: Advanced CRUD

- In addition to integer `ID`s that are auto-incrementing primary keys, the following cases should be supported:
  - `ID`s that are given in the payload on create, these might also be strings (enabled when the column does not equal the `primaryKey` in the provider)
  - `ID`s that are generated by a database function, these might also be strings (for example add something like this to the `sourcePath` of the feature: `{inserts=id=gen_random_uuid()}`)
- Adding additional columns with constant values on create should be supported (for example add something like this to the `sourcePath` of the feature: `{inserts=featuretype='type1'}`)
- Support provider mappings that make use of JSON columns 
- Support provider mappings that make use of joins

### Describe the Use Case

CRUD with more complex database schemas and mappings.

_No response_

### Additional Information

_No response_

azahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #1408.

Depends on interactive-instruments/xtraplatform-spatial#357

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '6.4.0-SNAPSHOT'
 xtraplatform-native  = '2.5.0'
-xtraplatform-spatial = '7.4.0-SNAPSHOT'
+xtraplatform-spatial = '7.4.0-crud-advanced2-SNAPSHOT'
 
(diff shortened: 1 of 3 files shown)


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1409.surge.sh

PR #1410, opened 2025-06-25
Upgrade JSON-FG, refactor support for profiles, add additional profiles

### LINKED ISSUE #521: JSON-FG: converting geometries to WGS 84
A follow-up from #519.

In order to support JSON-FG, GeoJsonWriterGeometry currently has special code to convert geometries that are not in WGS 84 back to WGS 84 and include it in the "geometry" member from GeoJSON. This code can currently only be triggered by the JSON-FG module.

Coordinate conversions should be handled in the transformation pipeline. However, before we update the transformation pipeline to support such a capability, we should wait, if JSON-FG will continue with the proposed GeoJSON-fallback with "geometry" always in WGS 84. 

This issue should be resolved before the JSON-FG module can eventually move into "stable".

### LINKED ISSUE #1397: Upgrade JSON-FG to v0.3

Link: https://github.com/opengeospatial/ogc-feat-geo-json/releases/tag/v0.3.0

### Describe the Use Case

This is the version of JSON-FG submitted to the OGC TC for the approval process.

Main changes:

- `featureType` is always a string (arrays are no longer allowed);
- Update conformance class URIs;
- Support also `\http://www.opengis.net/def/crs/OGC/0/CRS84` as a URI for CRS84;
- Use the GeoJSON media type for JSON-FG;
- Support the GeoJSON profiles.

### Additional Information

Support for CircularString plus CompoundCurve, MultiCurve, CurvePolygon, MultiSurface will be added through a separate issue.

### LINKED ISSUE #1419: Refactor support for profiles

OGC API standards increasingly specify profiles as variations of a media type, for example, to provide options how to represent feature properties with references to other resources or codelist values in the response. Another example are profiles that determine whether JSON-FG extensions are included in a GeoJSON response or not.

So far, profiles have been specified for feature and schema resources. 

To request one or more profiles, a query parameter "profile" can be used on resources that are associated with profiles.

The supported profiles depend on the resource type and, in some cases, on the media type (see the GeoJSON example above).

Profiles should be organized in profile sets. Whether a profile set applies to a resource representation depends on the resource type and potentially the media type of the representation. From each applicable profile set, at most one profile applies to the resource representation.

The profiles should be negotiated as follows:

- If a profile is specified via the "profile" query parameter and the profile is applicable to the resource, the profile is selected for the specific profile set;
- If no profile is explicitly requested for an applicable profile set, a default profile from the profile set is selected. Depending on the nature of the profile set, the default profiles can be specified either in the building block of the resource and/or in the building block of the format via the `defaultProfiles` option. A default specified for a format overrides the default profile specified for the resource.

The code should be updated to use the same patterns for identifying and negotiating applicable profiles.

The OGC Link object now also includes a "profile" link attribute (an array of profiles) and this information should be included in "self" links. For selected profile sets it should be possible to add "alternate" links not only for other media types, but also other profiles. Finally, a "profile" link has to be included for each profile.

In addition, the profile sets and their profiles should be documented more clearly, typically in a separate building block for a profile set. 

### Describe the Use Case

Simplify how new profiles can be added in the future and improve the documentation of profiles.

_No response_

### Additional Information

_No response_

### LINKED ISSUE #1420: Support profile for codelists in schemas, consolidate schema classes

OGC API - Features - Part 5: Schemas specifies two profiles for representing codelist values in schemas (as "enum" with a link to the codelist resource; as a one-of with one schema per code). Both should be supported.

### Describe the Use Case

Support the new standard profiles.

Add a new profile set based on #1419.

Currently, different classes are used for schemas that represent returnables/receivables vs queryables/sortables. However, they are mostly the same with small differences. These classes should be consolidated into a single set of classes for schema resources.

### Additional Information

_No response_

### LINKED ISSUE #1421: JSON-FG: Support embedding

#1280 added support for embedding feature references for GeoJSON. Add support for embedding referenced features also in the JSON-FG profiles of GeoJSON.

### Describe the Use Case

Using JSON-FG extensions together with embedded features.

_No response_

### Additional Information

_No response_

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1397.
Closes #521.
Closes #1419.
Closes #1420.
Closes #1421.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/361.

As part of the changes to JSON-FG, including support for embedding in all writers, the logic in the `GeoJSONWriter` classes has been updated. The stack of the features (root feature and embedded features) is now in the feature transformation context. The properties are always now buffered until all other applicable members (except links) have been written.

There are breaking changes to some existing API configurations:

- APIs that support the JSON-FG+ profile (previously: GeoJSON compatibility mode) now have to set the role `PRIMARY_GEOMETRY` to the geometry property for the "place" member. For the fallback geometry in the "geometry" member, a separate geometry property with the role `SECONDARY_GEOMETRY` has to be specified (it can use the same source path / geometry).

JSON-FG v0.3 also includes breaking changes to existing APIs based on changes in the candidate standard, including:

- There is no longer a plan to register a JSON-FG media type. The GeoJSON media type is now used and three profiles control the details of the representation.
- CURIEs are no longer supported.
- If "place" and "time" have no value, they are omitted instead of writing a `null` value.

Note: Tested with the "embed", "crud", "gvd" and "demo" configurations.


cportele:
I have updated the HTML schema view to include additional information from the JSON Schema (role, codelist, referenced collection) plus some fixes.

I noticed still one open issue in the schemas. The order of the properties is completely lost after the processing and properties appear in arbitrary order. This is quite confusing.

azahnen:
@cportele 
- I will add the PR Todo Checker that we already have in xtraplatform-spatial. I saw some TODOs during the review, especially the disabled tests might have to be addressed.
- I made various compare tests to check the GeoJSON is unchanged and JSON-FG only has the expected changes.
- I testet the feature profiles, still work as expected.
- I tested the schema profiles, there seem to be some issues:
  -  the schema still uses `x-ldproxy-codelistUri`, I think it should be `x-ogc-codelistUri`
  - `codelists-inline` has no effect, I see warnings in the log 
    `Codelist with ID 'cologne_lod2/RoofType' not found, returning schema without codelist information`
    http://localhost:7080/cologne_lod2/collections/building/schema?f=json&profile=codelists-inline

cportele:
@azahnen 

Thanks for the review. I have made the following changes:

- The problem with the codelist access in `cologne_lod2` has been fixed (the cause was that codelists in sub-folders was not handled properly).
- I have restored the "old" `featureType` configuration option, deprecated it, added `featureTypeV1` and a migration function to effectively use `featureTypeV1` only.
- I fixed the use of "x-ldproxy-codelistUri". I think this was still using "ldproxy" as at the time when I made the first changes, the codelist profile was not yet specified in Part 5.
- I have enabled the ignored unit test and updated it. I have also added a few more, but I have also removed some ignored tests that were really outdated. As GeoJSON and JSON-FG are key formats, we probably should add a follow-up task to add tests with more coverage (including embedded features, etc).
- The other TODOs were also reviewed (in general, these were not new).

azahnen (approved):
Looks good to me now.
For the TODOs you want to merge, you have to check the `Ignore` boxes. I think the bot will then resolve the conversations.

PR #1414, opened 2025-06-27
GeoJSON/JSON-FG: handle unsupported geometries

### LINKED ISSUE #1413: GeoJSON/JSON-FG: Improve handling of unsupported geometry types
### ldproxy Version

4.3.2

### Current Behavior

If an unsupported Simple Feature geometry type is provided as a geometry, an invalid GeoJSON / JSON-FG geometry object is written. 

### Expected Behavior

These situations should be detected, reported in the log and a `null` geometry should be written.

### Steps to Reproduce

Provide a geometry that is an empty geometry or a geometry collection.

### Relevant log output

```shell

```

cportele:
### Pull request checklist

-   [ ] Updated documentation

In the writers for "geometry" and "place" detect unsupported geometries. Write a `null` geometry instead and add a log for the first instance of an unsupported geometry type.

JSON-FG geometry types: Remove GENERIC (unused), map GEOMETRY_COLLECTION and add new method `isSupported()`.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/363

Closes #1413

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '6.4.0-SNAPSHOT'
 xtraplatform-native  = '2.5.0'
-xtraplatform-spatial = '7.4.0-SNAPSHOT'
+xtraplatform-spatial = '7.4.0-issue-1413-SNAPSHOT'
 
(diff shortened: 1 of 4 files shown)


azahnen (approved):
LGTM

PR #1447, opened 2025-08-13
Style tms

### LINKED ISSUE #1446: Styles: adjust MapLibre styles for other tile matrix sets

Add the capability to automatically derive MapLibre styles appropriate for tile matrix sets other than WebMercatorQuad from styles made for WebMercatorQuad. 

This avoid the need to maintain multiple copies of mostly redundant styles.

### Describe the Use Case

MapLibre styles are tied to tile sets in the WebMercatorQuad tiling scheme.

ldproxy supports tile sets in other tiling schemes, too. While MapLibre does not support the use of those tile sets, other clients can, including OpenLayers. 

To display such tile sets, some adjustments to the style are needed so the adjusted style can be used with tile sets in another tiling scheme. The adjustments are:

- Vector and raster sources need to be changed to tile sets for the other tiling scheme;
- Zoom levels may need to be adjusted, if the scale is not identical for each level of the WebMercatorQuad and the other tiling scheme.

A query parameter `tile-matrix-set` should be added for Style endpoints. The values are the tile matrix set IDs for tiles sets for the dataset or collection that are supported for style adjustments.

Tile matrix sets are supported, if they represent a Quadtree like WebMercatorQuad starting with a single tile on level 0. This includes, for example, AdV_25832 and AdV_25833.

In addition, the appropriate MapLibre styles should be used if OpenLayers is the `mapClientType` (using the [ol-mapbox-style](https://github.com/openlayers/ol-mapbox-style)) plugin.

### Additional Information

_No response_

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1446.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/383.

Still to do:

- [ ] update the OpenLayers map client
- [ ] identify a good sample style, test it and add a unit test for the derivation

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '6.5.0-SNAPSHOT'
 xtraplatform-native  = '2.5.0'
-xtraplatform-spatial = '7.5.0-SNAPSHOT'
+xtraplatform-spatial = '7.5.0-style-tms-SNAPSHOT'
 
--- a/ogcapi-draft/ogcapi-styles/build.gradle
+++ b/ogcapi-draft/ogcapi-styles/build.gradle
@@ -13,6 +13,7 @@ dependencies {
     provided 'de.interactive_instruments:xtraplatform-codelists'
     provided 'de.interactive_instruments:xtraplatform-crs'
     provided 'de.interactive_instruments:xtraplatform-features'
+    provided 'de.interactive_instruments:xtraplatform-tiles'
     provided 'de.interactive_instruments:ogcapi-collections'
     provided 'de.interactive_instruments:ogcapi-common'
     provided 'de.interactive_instruments:ogcapi-features-core'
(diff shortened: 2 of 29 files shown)


azahnen:
Deploying preview to https://ldproxy-ldproxy-pr-1447.surge.sh

cportele:
@azahnen 
FYI: The preview deployment failed. I don't see any error in the log, just "exit status 1".

cportele:
Thanks @PZahnen 👍 - I will have a look in the next days.

cportele:
@PZahnen

Regarding "The style for `AdV_25832` looks a bit unusual": If the tile matrix set is not `WebMercatorQuad`, the query parameter `tile-matrix-set` needs to be added to the style URL/request (e.g. `http://localhost:7080/strassen/styles/default?f=mbs&tile-matrix-set=AdV_25832`).

cportele:
@PZahnen 

I did a clean/rebuild and it now accessed the style for the tile matrix set (maybe it also did it before and I looked at the first request for the WebMercatorQuad tiles). The issue was that the resolutions for the levels in the tile matrix set was not passed to the `stylefunction` request.

Another issue was that the name of the first source was used, which may be another source, eg for the basemap. Then the style was not applied. See, e.g., in the daraa styles. I have changed this to select the first "vector" source.

PZahnen:
@cportele 
I see, thanks for the fix and the explanation 👍

azahnen (approved):
I implemented the additional links. If that looks good to you, please merge.

PR #1522, opened 2025-11-28
OpenLayers background map from style

### LINKED ISSUE #1520: OpenLayers client always uses OSM background map
### ldproxy Version

4.5.1

### Current Behavior

With `mapClientType: OPEN_LAYERS`, the background map is always OpenStreetMap, even if a different background map is set in the configuration or the MapLibre Style.

### Expected Behavior

Since OpenLayers uses the MapLibre Style to render vector layers (#1446), it should also use the background map defined in the MapLibre Style, if any.

### Steps to Reproduce

See above

### Relevant log output

```shell

```

### LINKED ISSUE #1521: add metadata to MapLibre styles in other tile matrix sets

As a follow-up to #1446, when a style is requested in other tile matrix sets, the style should contain a metadata element (e.g. `ldproxy:tileMatrixSets`) that contains an entry for each source with the corresponding tile matrix set.

### Describe the Use Case

This is necessary for example when a background map that only supports WebMercatorQuad is used. When that information is available in the metadata, clients may consider it for rendering

_No response_

### Additional Information

_No response_

azahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #1520.
Closes #1521.

### DIFF
--- a/ogcapi-custom/ogcapi-additional-metadata/src/main/java/de/ii/ogcapi/additional/metadata/domain/AdditionalMetadataConfiguration.java
+++ b/ogcapi-custom/ogcapi-additional-metadata/src/main/java/de/ii/ogcapi/additional/metadata/domain/AdditionalMetadataConfiguration.java
@@ -16,6 +16,11 @@ import org.immutables.value.Value;
 
 /**
  * @buildingBlock ADDITIONAL_METADATA
+ * @examplesAll <code>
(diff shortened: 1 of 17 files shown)


azahnen:
Deployment of preview was successful: https://ldproxy-ldproxy-pr-1522.surge.sh

PR #1528, opened 2025-12-07
show feature title instead of id in items map popup

azahnen:
### Pull request checklist

-   [ ] Updated documentation

The popup/popover in the map on the items page currently shows the id, which in most cases is not helpful for a user. This PR changes the popup to contain the title instead that is also used as heading in the list and can be configured using `featureTitleTemplate`.


azahnen:
Deployment of preview was successful: https://ldproxy-ldproxy-pr-1528.surge.sh

PR #1530, opened 2025-12-09
glTF: add embedSchema and schemaUri options

### LINKED ISSUE #1529: glTF: add options to declare the EXT_structural_metadata schema

The glTF extension `EXT_structural_metadata` needs a schema that describes the properties. This can be embedded in the glTF asset or referenced. Currently the schema is always referenced using the URI of the schema as published via the API.

There should be options to control the representation of the schema. It should be possible to embed to schema or reference the schema using a pre-defined URI outside of the API.

### Describe the Use Case

A key use case where these options are needed is to seed the subtree and content files of a 3D Tiles tileset in a different environment to the environment where the tileset will eventually be deployed.

Add two new options in the glTF building block:

- `embedSchema`, default `false`: If set to `true`, the glTF schema describing the feature properties as part of `EXT_structural_metadata` extension is embedded in the glTF asset instead of referencing the schema by URI.
- `schemaUri`, default `null`: By default, the glTF schema describing the feature properties as part of the `EXT_structural_metadata` extension is referenced by the URI under which the schema is published via the API. Use this option to specify a different URI. Note that the option is ignored, if `embedSchema` is set to `true`.

### Additional Information

_No response_

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1529

### DIFF
--- a/ogcapi-draft/ogcapi-features-gltf/src/main/java/de/ii/ogcapi/features/gltf/app/FeatureEncoderGltf.java
+++ b/ogcapi-draft/ogcapi-features-gltf/src/main/java/de/ii/ogcapi/features/gltf/app/FeatureEncoderGltf.java
@@ -605,16 +605,28 @@ public class FeatureEncoderGltf extends FeatureObjectEncoder<PropertyGltf, Featu
 
     PropertyTable propertyTable = propertyTableBuilder.build();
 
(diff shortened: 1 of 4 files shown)


azahnen:
Deployment of preview was successful: https://ldproxy-ldproxy-pr-1530.surge.sh

PR #1593, opened 2026-04-22
Features in HTML: update CRS handling

### LINKED ISSUE #1585: Features in HTML: CRS handling results in incorrect responses
### ldproxy Version

4.7.0

### Current Behavior

This is a follow-up to https://github.com/ldproxy/ldproxy/issues/1561, where the CRS Editor was suppressed when Cesium is used as the map client in a HTML representation of features.

A closer look shows that there are also issues in all cases, i.e., also where MapLibre is the map client: The coordinates in the schema.org annotations have to be in WGS84 lat/lon, but they are in the requested CRS with swapped XY axes.

### Expected Behavior

Both the code for the schema.org annotations and for the map client expects that the feature geometries are in the default CRS. The feature geometries are not used for anything else in the HTML representation. 

That is, the `crs` parameter is not relevant for the HTML representation of features and should be ignored in that case when querying the features. The `crs` parameter, however, still has to be used in all links, in particular the `self`/`alternate` links.

If the code is changed so that the `crs` parameter is ignored in the generation of the feature query, if the target representation is HTML (in addition to dropping the `crs` parameter in the feature request by the map client), the schema.org annotations will be correct, the links will be correct and the Cesium map client will also shows the features, if the `crs` query parameter is specified in the URL. 

That is, the limitation that the CRS editor is disabled, if Cesium is the map client, could be removed, too.

### Steps to Reproduce

Request feature with `f=html` and `crs` with a CRS that is not the default CRS. Inspect the schema.org annotation in the HTML. In addition, if Cesium is the map client, it will not display the features.

### Relevant log output

```shell

```

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1585.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1593.surge.sh

azahnen (approved):

PR #1594, opened 2026-04-23
custom cql2 functions

### LINKED ISSUE #1591: CQL2 custom functions

Support the definition of custom CQL2 functions in SQL feature provider configurations.

### Describe the Use Case

Support user defined functions in CQL2 filter expressions.

_No response_

### Additional Information

Configuration example:

```yml
cql2Functions:
  - name: MY_UPPER_MATCH
    arguments:
      - name: prop1
        type: [STRING]
      - name: prop2
        type: [STRING]
    returns: 
      - BOOLEAN
    expression: 
      SQL: "UPPER($prop1) LIKE UPPER($prop2)"
```

PZahnen:
### Pull request checklist

-   [x] Updated documentation

Closes #1591 .

Depends on ldproxy/xtraplatform-spatial#482


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1594.surge.sh

cportele:
@azahnen @PZahnen 

Some observations after reviewing and testing:

- The documentation of the building block FILTER should be updated. A reference to the configuration of custom CQL2 functions should be added (it took me a while to figure out where they are configured).
- In the scope section of the FILTER building block there is a reference to a function `NOW()` which is not in the list of functions, so I guess this is another case that has special code. It would be good to get rid of that special scope, too.
- I see that the arguments on `/functions` cannot be an array of values based on OGC API Features Part 3. I will create an issue in OGC API Features about that.
- It would be good to be able to add a description to each argument (OGC API Features Part 3 supports it and it will be helpful to users/AI models). We could also do this later.
- When I configure a custom function, but do not include an expression for the feature provider type, the CQL2 expression is still accepted and converted into a SQL statement which uses the custom CQL2 as if it would be a SQL function. There is no error (there should be one, or probably only custom functions with an expression for the provider should be included in the set of custom functions), the request simply never finishes with no error in the log.
- I noticed the previous issue when I copied a custom function to another provider that uses a different database type, but I did not update the key in the expression map. Since the custom functions are defined with a provider, in the standard case the expression as a map does add anything, its is just easier to make errors in the configuration. I guess the map in "expression" is for the case where the custom functions are defined in "defaults"?

azahnen:
@cportele Regarding the last point, yes, I assumed that in many cases functions would be defined per deployment and not per API. We could add a shortcut when functions are defined in the provider, e.g. `expression` and the map in `expressions`? And then of course implement error handling for your previous point that also handles these cases.

cportele:
@azahnen - sounds good to me

PZahnen:
Thanks for the review! Here's how the individual points were addressed:

Point 1: The documentation for the FILTER building block has been updated to include a reference to the configuration of custom CQL2 functions.

Point 2: Previously, NOW() was handled as an implicit special case in CqlTextVisitor, expanding it inline to Instant.now(). It is now consistently delegated to TemporalLiteral.now() and additionally registered as a built-in function in CqlBuiltInFunctions, making it visible in the central function list.

Point 4: Support for argument descriptions was already integrated; it has now also been explicitly documented.

Points 5 & 6:  Implemented as described above.

cportele:
Thanks @PZahnen 

There was an issue in the links (fixed).

There is one oustanding issue that - as far as I can see - has not yet been addressed: If `expressions` does not include an expression for a particular feature provider (e.g., `SQL/PGIS` is specified, but the feature provider is a GeoPackage provider), no error is written to the log during startup and the function is still available in the feature provider (but results in a request that never finishes). That function should only be available in PostGIS feature providers and disabled for all other feature providers. Or, if that is not possible, the use of the function should result in an error.

PZahnen:
Thanks for the follow-up @cportele 

We now validate custom CQL2 function expressions against the active provider SQL dialect during startup. If no matching expression exists, the function is excluded for that provider and an ERROR is logged. Calling such a function now returns `400 Unknown function` (no hanging request).

cportele (dismissed):
Looks good now, thanks

azahnen (approved):

PR #1596, opened 2026-04-24
fix conversion to JSON Schema

### LINKED ISSUE #1595: Incorrect mapping of a feature schema with nested objects to JSON Schema
### ldproxy Version

4.7.0

### Current Behavior

The ldproxy feature schema provides `label` and `description` annotation for a feature type and for its properties. In case of object-valued properties the properties of that object are represented as nested properties. That is, there is no support for a label or description of a data type - only the type name can be provided with the property in the `objectType` annotation.

When the feature schema is transformed into a JSON Schema, the `label` and `description` of the property are moved to the object type in the `$defs` section and the property has no `label` or `description`.

### Expected Behavior

The `label` and `description` of nested object types under `$defs` should be omitted and instead be represented with the property.

### Steps to Reproduce

Take any collection with an `OBJECT` or `OBJECT_ARRAY` property in its feature schema, enable the `SCHEMA` module and request the schema for that collection.

### Relevant log output

```shell

```

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1595


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1596.surge.sh

azahnen (approved):

PR #1619, opened 2026-05-13
NAS support

cportele:
### Pull request checklist

-   [x] Added/updated unit tests
-   [x] Updated documentation
-   [x] All checks are passing

### Changes introduced by this PR

Closes #1618. 

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/502.

PR #1622, opened 2026-05-18
add GML support for CRUD

### LINKED ISSUE #1621: add GML support for CRUD

Follow-up to #1618. Add the reverse direction: a ldproxy instance that already serves GML data through the OGC API should accept GML bodies on POST and PUT and persist them through the existing SQL feature-encoder pipeline. In particular, NAS input should be supported. 

- `application/gml+xml` should be accepted as a request body content type for POST (create) and PUT (create or replace) on collections that have the CRUD building block enabled and the GML building block configured. PATCH stays JSON-only and rejects GML bodies with HTTP 415.
- The body must be a single bare feature element at the document root; multi-feature bodies (e.g. `wfs:FeatureCollection`) are rejected with HTTP 400.
- The full set of `GmlConfiguration` options that describe a feature type's GML shape (`applicationNamespaces`, `defaultNamespace`, `objectTypeNamespaces`, `useAlias`, `gmlIdPrefix`, `srsNameMappings`, `uomMappings`, `codelistProperties`, `codelistUriTemplate`, `featureRefTemplate`, `valueWrap`, `xmlAttributes`, `variableObjectElementNames`) should be honoured symmetrically by the decoder. The remaining options stay output-only.                                                                                                 
- The coordinate reference system of each geometry should be resolved in this order: the geometry's own `srsName` attribute, the `Content-Crs` request header, the collection's storage CRS. A feature whose geometries mix coordinate reference systems is rejected with HTTP 400. Community-specific URIs on srsName (e.g. `urn:adv:crs:ETRS89_UTM32`) are resolved through `GmlConfiguration.srsNameMappings`.
- When the request carries `Prefer: handling=strict`, the body is validated against the XSDs referenced by `GmlConfiguration.schemaLocations` (in addition to the structural checks the decoder performs unconditionally). Missing or unparseable schemas log a warning and skip validation rather than failing the request.

### Describe the Use Case

Accept NAS/GML features in mutation requests.

The work spans both `xtraplatform-spatial` and `ldproxy`.
                                                          
`xtraplatform-spatial` should gain the GML write-path foundation:
  
- A new schema-resolved decoder (`FeatureTokenDecoderGml`) walks an incoming GML feature against the feature schema and emits values at the dotted property-name path that the SQL feature encoder's writable/value column maps already consume. Supports nested OBJECT(_ARRAY), FEATURE_REF(_ARRAY), GML array properties (one property element wrapping multiple peer object elements), per-level value/feature-ref/object-array bracketing at every container depth, geometry decoding at any nesting depth, `valueWrap` chains with hardcoded `gmd`/`gco` value-wrapper conventions, `xsi:nil` → `null`, `variableObjectElementNames` on the feature root and on nested objects, namespace-aware property matching, and an optional `wfs:FeatureCollection/member` wrapper chain (a multi-feature shape is rejected, but the configured wrapper around a single feature is accepted).
- A new `FeatureTokenDecoderGmlInputProfile` defines the reverse mappings the decoder needs, alongside the existing encoder profile. ldproxy derives an instance from `GmlConfiguration` at format-extension wiring time.
- The existing WFS-shape decoder is renamed from `FeatureTokenDecoderGml` to `FeatureTokenDecoderGmlFromWfs` so the path-shape distinction is explicit at the type level; the new schema-resolved decoder takes the unsuffixed name.
- `GeometryDecoderGml` is rewritten on a stack-based context model so it covers the GML 3.2 geometries used by NAS (Point/MultiPoint, LineString/MultiLineString, Curve and CompositeCurve and MultiCurve with LineStringSegment/Arc/ArcString/Circle, Polygon and CurvePolygon, Surface/CompositeSurface/PolyhedralSurface with PolygonPatch, Solid as closed PolyhedralSurface, GeometryCollection). Constructs outside this list should be rejected with a clear error rather than producing silently wrong geometry. CRS inheritance from outer to inner geometry frames is preserved so a resolved CRS on an enclosing CompositeCurve survives down to inner LineStringSegments.
- `GeometryDecoderGml` accepts an `srsNameMappings` parameter and consults it before the built-in EPSG / OGC URN parsers, so application-profile srsName URNs resolve.
- A unit-level fix in the SQL pipeline: FeatureEncoderSql now opens and closes a junction-table row around each VALUE_ARRAY element so the junction-table writable column (which existed in the mapping but was never reached) is actually written; SqlMappingDeriver skips object-table dedup for VALUE_ARRAY tables so the cleanup pass does not drop their value column.
- Bare-feature ALKIS NAS sample documents (Datenlizenz Deutschland Zero 2.0, City of Bonn) drive new Spock specs that cover both the decoder and the geometry decoder round-trip.

In `ldproxy`:

- `EndpointCrud` passes the request MediaType through to the command handler in place of the prior `jsonFg` boolean, and forwards the `Link` header list to allow for profile detection. `CommandHandlerCrudImpl` no longer carries the JSON Schema validation pipeline; instead it builds a format-agnostic `ValidatorContext` (api/collection/media type, request context, profiles parsed from `Link` headers, default schema-validation profiles) and dispatches to `FeatureFormatExtension.validate(content, ValidatorContext)`. The relocated schema cache is split by scope (`ReceivablesJsonSchemaCache`, `ReturnablesJsonSchemaCache`) and moved out of the CRUD module.
- `FeaturesFormatGeoJson.validate` keeps the existing JSON-Schema validation behaviour, now resolving the right profile (returnables/receivables × GeoJSON/JSON-FG) from the declared profiles on the request.
- `FeaturesFormatGml.validate` adds XSD validation of the request body against the XSDs referenced from `GmlConfiguration.schemaLocations`.
- The `CrudBuildingBlock` and `GmlConfiguration` JavaDoc are updated to describe the new content-type matrix, the bare-feature requirement, the CRS resolution order, and the per-option direction notes (which `GmlConfiguration` options are honoured on input vs. output-only).                                                                        
- Live smoke specs `TransactionalRESTApiSpec` (GeoJSON) and `TransactionalGmlRESTApiSpec` (GML) are resurrected on `java.net.http.HttpClient` so they work under Groovy 4 and self-adapt to the collection's `supportsNonAutogeneratedResourceIds` flag.

### Additional Information

_No response_

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1621.

Depends on #1619 and https://github.com/ldproxy/xtraplatform-spatial/pull/507.

Also includes two fixes:
- In the SQL pipeline: `FeatureEncoderSql` now opens and closes a junction-table row around each VALUE_ARRAY element so the junction-table writable column (which existed in the mapping but was never reached) is actually written; `SqlMappingDeriver` skips object-table dedup for VALUE_ARRAY tables so the cleanup pass does not drop their value column.
- A single-feature GET on a missing feature with `f=xml` now returns 404 (it previously returned 500 because the GML encoder tried to close a document that never opened a root element).


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1622.surge.sh

azahnen (dismissed):
only reviewed design, looks good to me

PR #1627, opened 2026-05-24
support transactions

cportele:
### Pull request checklist

-   [x] Added/updated unit tests
-   [x] Updated documentation
-   [x] All checks are passing

### Changes introduced by this PR

Closes #1623.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/512 and #1622.

Performance test with inserting 10 000 AX_Flurstueck features:

 | Format                              | No validation | `handling=strict` |
 | ----------------------------------- | ------------- | ----------------- |
 | `application/ogc-tx+json`  with GeoJSON         | 3.7 s        | 4.0 s            |
 | `application/xml` (wfs:Transaction with GML) | 24.4 s        | 27.2 s            |

PR #1634, opened 2026-06-06
transaction and GML improvements, support versioned features

### LINKED ISSUE #1640: add support for OGC API Features: Versioned Features

A proposal for a new part [OGC API Features: Versioned Features](https://github.com/opengeospatial/ogcapi-features/blob/master/proposals/versioned-features.md) has recently been added and is being discussed in OGC. The new proposal should be supported.

### Describe the Use Case

Datasets like cadastral or road data sometimes represent the changes in the real-world over time, often using a Versioned Features appraoch. That is, they allow querying the data at arbitrary times or allow analyzing the changes over time. It would be good to be able to support this with appropriate API components.

- Implement the proposal.
- For mutations, support Transactions, but skip CRUD for now. This can be added later.
- Do not support the `features-with-time-slices` profile.
- Also support dates as the time-granularity. Test with the CShapes dataset used on demo.ldproxy.net. Since the temporal extent of the dataset ends in 2019, a different default than "now" should be supported in the `datetime` query parameter.
- For 410-GONE responses, in an HTML response also return a proper page with links to the Time Map and the latest version.

### Additional Information

_No response_

cportele:
# Pull request checklist

-   [x] Updated documentation

# Changes introduced by this PR

Follow-up to #1627. Closes #1640.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/525.

## Transactions

- Improves "Update" actions to native SQL UPDATEs on the session's own connection, gated by a per-collection whitelist (new configuration option `updatableProperties`, limited to VALUE, GEOMETRY, VALUE_ARRAY and OBJECT_ARRAY without nested objects or arrays). Avoids the GET-merge-write update path to improve performance and supports multiple actions on the same feature within one atomic transaction.
- Update supports property paths.
- Failure reporting identifies the failing feature(s) in both the server log line and the ActionResult, and limits stack traces to system / infrastructure errors.
- The endpoint and its handlers report their health (`ApiExtensionHealth`, volatile composition).

## Versioned Features

New `VERSIONED_FEATURES` building block (maturity PROPOSAL, registered in ldproxy-cfg) for collections where each version of a feature is a separate row and the primary temporal interval bounds its validity.

### Mutations

- Per-action behaviour for versioned collections, dispatched via the MutationStrategy; every flow is whitelisted by the building block.
- The open version's pointer columns use the PREDECESSOR_INTERVAL_START / SUCCESSOR_INTERVAL_START roles; the executor surfaces the retired version's start via insertRoleOverrides so the new row's predecessor column lands populated, and retireFeature
writes the matching successor pointer on the closed row.
- Composite ids (feature id + version start, pattern and timestamp format per collection configurable) in replace/update/delete action filters carry an If-Unmodified-Since-style expectedStart predicate end-to-end; a stale composite id yields 412
Precondition Failed.
- Update mode dispatch: native patch-in-place for non-versioned collections, retire-in-place when the update only sets PRIMARY_INTERVAL_END (optionally with whitelisted siblings via `retireWithModifications`), clone-and-patch when the update creates a new version.

### Queries

- The `datetime` parameter on `items/{featureId}` defaults to `now` — configurable per collection via `defaultDatetime` — and selects the version valid at that instant; intervals are rejected on the single-feature resource.
- New Time Map resource `/collections/{collectionId}/items/{featureId}/versions`: every version as an RFC 7089 `memento` link with a `datetime` link attribute, plus `original` and `latest-version` links; JSON, HTML and `application/link-format` (RFC
6690) representations.
- Single-feature responses carry `predecessor-version` / `successor-version` links as HTTP `Link` headers and per-format body content (GeoJSON/JSON-FG links, HTML version navigation, GML comment convention).
- Profiles `versions-as-features` and `versions-as-features-unique-ids` for representing all versions on `items`; the latter appends the version start to the feature id.
- `410 Gone` instead of `404` when the feature exists but no version matches the requested datetime, with `original` / `latest-version` / `timemap` links, an HTML page and problem+json.
- DATE-typed primary intervals are fully supported and dates stay dates: memento hrefs, HTML labels, composite-id suffixes (`yyyyMMdd` default) and the 410 message use the date value, not a derived timestamp.

### Properties as web links

API-side resolution of schema-configured property links (see the xtraplatform-spatial PR): the URI template parameters `{{value}}`, `{{featureUri}}`, `{{collectionUri}}` and `{{serviceUri}}` are resolved against the request URIs. Link properties
surface as HTTP `Link` headers on single-feature responses and as link entries in the formats; the previously hard-wired versioning links are now a special case of this mechanism.

## GML improvements

(Could also be a separate PR as they are not dependent on the transaction changes, but are included as they had to be tested together.)

- Improves XSD validation with a local XSD catalog (faster, more reliable; an `xsdCatalog` map on the GML building block, resources under `<data>/resources/xsdCatalog/`; the validator's `LSResourceResolver` consults the catalog first for nested imports) and caches Schema/Validator objects.
- For GML with mixed namespaces, the original object type of each property (new `originObjectType` schema member, see the xtraplatform-spatial PR) drives namespace-correct encoding and decoding of properties.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1634.surge.sh

cportele:
@azahnen - this should now be ready for a review.

The changes have been tested with NBA transactions including insert, replace and update statements on the same feature in a single transaction. The query path has been tested against that data, too, as well as against the CShapes demo API with a DATE-based interval. https.//demo.ldproxy.net/cshapes has been updated.

We should check in a follow-up task, if the code would benefit from more refactoring; e.g., to consolidate the SQL statement mutation templates.

CRUD has not been updated yet for Versioned Features. This will be left for a follow-up task and we first need to discuss some design decisions.

An issue for Versioned Features (Queries, Transactions) has to be created before merging the PR.

azahnen (dismissed):
No notes.

PR #1647, opened 2026-06-16
Search: result sets, deduplication, and geometry parameters in query expressions

### LINKED ISSUE #1646: Search: chain queries via semi-joins in a single query expression (result sets)

The SEARCH building block (OGC API – Features – Part 10: Query, draft) executes a query expression: one or more independent queries, each selecting features from one collection, optionally with a shared top-level `filter` / `filterOperator` and stored-query `$parameter` substitution. The queries are merged into one FeatureCollection.

What the model cannot express today is feeding one query's selection into another query's predicate. There is no way to say "select the features of collection B that are referenced by — or that reference — the features selected earlier from collection A", and then continue that into C, D, … This is a chain of semi-joins: each step filters one collection by its relationship to a set of features produced by an earlier step, while still emitting each collection's features into the merged response.

This is needed wherever a request must traverse association/reference chains across several feature types and return the features encountered along the way. This is similar to the `resolvePath` option in WFS 2.0 queries.

A handful of supporting gaps surface together with this:

- Features selected by more than one query (different steps reaching the same feature) appear multiple times in the merged response; there is currently no opt-in deduplication.
- A step may need to contribute a set for later steps to consume without emitting its own features.
- Stored queries currently cannot take a geometry-valued parameter, and `filterCrs` cannot be parameterized — both needed to expose, e.g., an area-of-interest as a stored-query parameter.
- When a provider's feature ids are globally unique, the merged response should keep the original ids rather than qualifying them with the collection id (needed when ids must remain stable across the merge and when deduplicating on id alone).
- Query expressions inherit paging (`offset` parameter; `next` link) from single-collection queries, but a chained, multi-query response should probably be single-shot; paging across the chain is not well-defined. For large query results, the more appropriate design is to support asynchronous request execution, which will be a follow-up task.

### Describe the Use Case

See the description above.

**Result sets and inResultSet (SEARCH + CQL2)**

A query in a query expression may declare named result sets:

```json  
"resultSets": {
  "<name1>": {},                         // ids of the features selected by this query
  "<name2>": { "values": "<queryable>" } // ids referenced via a reference property of this query
}
```

- A plain set is the ids of the query's selected features.
- A projected set is the values of one FEATURE_REF / FEATURE_REF_ARRAY queryable of the query's own collection (validated there).
- Shorthand `"resultSet": "<name>"` ≡ `"resultSets": {"<name>": {}}`.

A later query references a set through a new CQL2 predicate:

```json
{ "op": "inResultSet", "args": [ { "property": "<p>" }, "<setName>" ] }
```
  
- `<p>` is a queryable of the collection.
- Semantics: `IN` for a single-valued property, set-overlaps for an array-valued one, against the set's ids.
- A forward step (relation stored on the producing side) is expressed as a projected set on the producer + `inResultSet(id, set)` on the consumer; the reverse step is a plain id set on the producer + `inResultSet(<refProperty>, set)` on the consumer. Both compile to a semi-join in the relation's storage direction — neither requires walking a relation against how it is stored.

Rules: set names are unique within a query expression; a query may reference only sets declared by earlier queries (forward-reference-free DAG, no cycles); declared-but-unused sets are allowed; sets are computed from the producing query's full selection (a response `limit` truncates output, never the sets); all collections of one query expression must be served by the same feature provider. For references with several target types, a set's ids only match where the set's value type is a valid target type of the reference.
  
**Supporting options**

- `resultSetOnly: true` on a query — it defines its sets but contributes no features and is excluded from `numberMatched`/`numberReturned`.
- `deduplicate: true` on the query expression (opt-in) — a feature selected by more than one query is emitted once. To be implemented as a streaming, format-agnostic token filter keyed on the feature id (or `collectionId.featureId` when ids are not globally unique), with a memory-bounded packed-id set. `numberMatched`/`numberReturned` are SQL-computed and may overcount dropped duplicates (to be documented as a limitation; an exact count would need a second pass). This may be changed in the future, of needed.
- `computeNumberMatched` toggle on the query expression — disabling it avoids one count query per sub-query.

**Other**

- Geometry-valued stored-query parameters (`{"format": "geometry-<type>"`}, accepting WKT or GeoJSON, with type enforcement) and a parameterizable `filterCrs`.
- Skippable id rewriting — the merged response keeps original ids automatically when the provider declares globally unique ids (provider option `globallyUniqueFeatureIds`).
- No paging for query expressions — `offset` and `next`/`prev` links are dropped; `limit` remains a hard cap. Large extracts are the use case for future asynchronous execution, not paging.

**Execution**

Result sets compile to SQL subqueries generated at encode time: a plain set to `SELECT <id> FROM <table> WHERE <predicate>`, a projected set to `SELECT <ref-column> FROM <table> [JOIN <junction>] WHERE <predicate>`, and `inResultSet(p, S)` to `p IN (<subquery of S>)` with the existing array/EXISTS handling; chains nest recursively.

Because a naïvely inlined chain re-derives a shared producer once per consumer (and re-evaluates a deep chain at every nesting level), large fan-outs should be optimized two ways:

1. Within a statement, each result set is hoisted into a `WITH … AS MATERIALIZED CTE` (dialect-controlled; PostgreSQL needs the explicit `MATERIALIZED`), so it is evaluated once.
2. Across statements, the result sets of a multi-query are materialized once per request — each producer runs a single time (its own dependencies already inlined as literal id lists), and the collected ids are attached to the `inResultSet` nodes so consumers  encode a literal `IN` list. Producers are grouped into dependency levels and the independent producers of a level run concurrently (bounded by the connection pool). A set larger than a configurable cap (`queryGeneration.resultSetMaterializationMaxSize`, default `100 000`) falls back to the inline subquery.

**Scope**

The extension spans two repositories — CQL2 + feature/SQL execution in `xtraplatform-spatial`, the query-expression model + handler in `ldproxy`. 

Out of scope: asynchronous execution for large extracts and per-query statistics.

### Additional Information

**Relationship to Part 10**

`resultSets` / `inResultSet` / `resultSetOnly` are an additive extension of the JSON query expression (two query members + one CQL2 predicate). They could be submitted as a candidate for a new requirements class in Part 10,
complementary to the existing capability for full joins.

cportele:
### Pull request checklist

-   [x] Updated documentation

Adds chained semi-joins to the SEARCH query expression. 

Closes #1646. 

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/533.

## Query-expression model (`ogcapi-features-search`)

- **`resultSets`** on a query (`ResultSetDefinition`, optional `values`) and the **`resultSet`** shorthand on `SingleQuery` / `SingleQueryWithParameters`, copied
  through `ParameterResolver`.
  - `{}` → ids of the query's selected features; `{ "values": "<queryable>" }` → ids referenced via a `FEATURE_REF`/`FEATURE_REF_ARRAY` queryable of the query's own collection.
- **`resultSetOnly: true`** on a query — defines its sets but emits no features and is excluded from the counts.
- **`deduplicate: true`** on the query expression (default false) — drops a feature selected by more than one query.
- **`computeNumberMatched`** on the query expression — disables the per-sub-query count.
- **`offset` removed** from the model; the corresponding query parameter and the `next`/`prev`/`first` links are gone. `limit` remains a hard cap.

### Worked example

Three collections — `building`, `address`, `parcel` — where each `building` references its `address` (a `FEATURE_REF` queryable `address`) and each `parcel` references the `building` it sits on (a `FEATURE_REF` queryable `building`). The request selects buildings in an area of interest, then the addresses they reference, then the parcels that reference those buildings — all merged into one `FeatureCollection`, deduplicated:

```jsonc
{
  "deduplicate": true,
  "queries": [
    {
      // seed: buildings in the area of interest; declare two result sets
      "collections": [ "building" ],
      "filter": { "op": "s_intersects", "args": [
        { "property": "geometry" },
        { "$parameter": { "$ref": "#/parameters/AreaOfInterest" } } ] },
      "resultSets": {
        "selected_buildings": {},                  // ids of the selected buildings
        "their_addresses": { "values": "address" } // ids referenced via `address`
      }
    },
    {
      // forward step: the referenced addresses match by their own id
      "collections": [ "address" ],
      "filter": { "op": "inResultSet",
                  "args": [ { "property": "id" }, "their_addresses" ] }
    },
    {
      // reverse step: parcels whose `building` reference points at a selected building
      "collections": [ "parcel" ],
      "filter": { "op": "inResultSet",
                  "args": [ { "property": "building" }, "selected_buildings" ] }
    }
  ]
}
```

A *forward* step (the relation is stored on the producing side) uses a projected set plus `inResultSet(id, set)` on the consumer; a *reverse* step uses a plain id set plus `inResultSet(<refProperty>, set)`. Both compile to a semi-join in the relation's storage direction. Set names are unique, a query may reference only sets declared by earlier queries, and all collections must be served by the same feature provider.

## Execution (`SearchQueriesHandlerImpl`)

- Builds the result-set registry in query order, resolving each `inResultSet` reference against the producing query (`ResultSetResolver`): 400 on a duplicate set name, an unknown reference, or a forward reference. `resultSetOnly` queries are kept in the registry but excluded from the `MultiFeatureQuery` and from the collection-id zipping that aligns types with collections.
- **Filter CRS** is now applied to query-expression filters (each effective filter is routed through `cql.read(cql.write(f), filterCrs)` before result-set resolution), and coordinates are validated when the API enables `validateCoordinatesInQueries` (invalid input → clean 400).
- **Skippable id rewriting**: the merged response keeps original feature ids when the provider declares globally unique ids, otherwise qualifies them with the collection id; also fixes per-feature links in multi-collection results.

## Stored-query parameters (`ParameterResolverCql`)

- **Geometry-valued parameters**: schema `{"format": "geometry-<type>"}`, value accepted as **WKT or GeoJSON** (GeoJSON detected by a leading `{`), decoded with `filterCrs` into a spatial literal; the geometry **type is enforced** post-decode (wrong type → 400). `filterCrs` itself is parameterizable.
- `OpenApiSchemaDeriver` maps the geometry parameter schema to a string schema with the geometry format for the generated OpenAPI definition.

## Validation

- Stored-query and query-expression filters are validated against the collection's queryables before execution (`StoredQueryValidator`), including filters that contain still-unbound parameters; fixes a defaulted filter parameter resolving to `null`.

## Documentation

Building-block JavaDoc (`@langEn`/`@langDe`) updated for result sets, `resultSetOnly`, deduplication, geometry parameters, and the removal of paging.

## Tests

`ResultSetSpec`, `FilterCrsSpec`, `ParameterResolverCqlSpec`, `OpenApiSchemaDeriverSpec` (parsing of `resultSets`/shorthand/`values`/`resultSetOnly`, the result-set resolver, filter-CRS handling, WKT/GeoJSON geometry parameters with type enforcement, OpenAPI derivation).


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1647.surge.sh

azahnen (dismissed):
Looks good to me, no testing.

PR #1648, opened 2026-06-17
HTML feature view: fix empty nested objects

cportele:
### Pull request checklist

-   [ ] Updated documentation
-   [X] All checks are passing

Object properties whose entire subtree resolves to no value (nested object properties with only null leaves) were rendered as bare labels in the HTML feature view. Adds a recursive `hasContent()` to `PropertyHtml`, excludes empty objects from `getObjects()`, and guards the object/array rendering branch in `property.mustache` with `hasContent` so empty objects self-suppress at any depth.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1648.surge.sh

azahnen (approved):

PR #1650, opened 2026-06-18
gml: encode feature-reference elements with an object-type name suffix

cportele:
### Pull request checklist

-   [x] Updated documentation

Follow-up to #1619.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/537.

#### Encode feature-reference elements with an object-type name suffix

Some GML application schemas name a feature-reference property element with the referenced feature type appended (e.g. `foo_Bar` and `foo_Baz` instead of `foo` as the relationship role is named in the logical model). 

In the GML feature encoder, emit `foo_<ObjectType>` instead of the plain element, with the suffix derived from the referenced collection's object type.

- `GmlConfiguration`: new `objectTypeSuffixedProperties` option.
- GML now encodes feature references natively instead of relying on the generic `rel` reduction: `FeaturesFormatGml.getPropertyTransformations` drops the `rel` profile set, so the resolved `id`/`title`/`type` object reaches the encoder, and `GmlWriterProperties` builds `xlink:href`/`xlink:title` and appends the `_<ObjectType>` suffix resolved from the type.
- `FeatureTransformationContextGml`: `REF_SUFFIX` placeholder, `refValues` collection, `refTargetObjectTypes`.
- `toInputProfile` forwards `objectTypeSuffixedProperties` to the decoder profile.
- Specs: `ToInputProfileSpec`, `WithoutRelProfilesSpec`.

#### Encode ISO 19139 codelist values as codeList/codeListValue

A second change has been added to this pull request as it also affects `GmlWriterProperties` to avoid merge conflicts later.

A scalar codelist value wrapped (via valueWrap) in an ISO 19139 codelist element was emitted bare, e.g. <gmd:CI_RoleCode>processor</gmd:CI_RoleCode>, which the application schema requires to carry the codeList and codeListValue attributes.

Add the GML option codeListUriTemplateIso19139. When set, a property that references a codelist (its codelist constraint in the provider schema) and is wrapped in an element whose local name equals the codelist id is encoded as an ISO 19139 codelist value: the wrapping element receives codeList (the template with {{codelistId}} substituted) and codeListValue (the value, which is also the element text). The option is opt-in (null disables it) and affects output only; the decoder reads such values from the element text and ignores the two attributes.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1650.surge.sh

azahnen (approved):

PR #1651, opened 2026-06-18
gml: fix namespace collisions

cportele:
### Pull request checklist

-   [ ] Updated documentation

Whenever a collection's GML configuration declared an applicationNamespaces prefix that also exists in the standard set (gml, gml21, gml31, xlink, xml, xsi, sf, wfs), an exception was thrown.

Fix: merge the two sets explicitly. The reserved prefixes always win, since the GML writers reference them internally; an application namespace that reuses a reserved prefix is dropped. A reserved prefix rebound to a different namespace is logged as a misconfiguration; reusing it for the identical namespace is dropped silently. The merged map is cached per (api, collection), so the merge and any warning happen once rather than on every encoded response.

Add MergeNamespacesSpec covering the empty, non-reserved, same-URI, and differing-URI cases.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1651.surge.sh

azahnen (approved):

PR #1659, opened 2026-06-23
wfs:FeatureCollection: numberMatched/numberReturned are required attributes

cportele:
### Pull request checklist

-   [ ] Updated documentation

The two XML attributes of `wfs:FeatureCollection` are required by the WFS 2.0 schema, but the attributes were not written, if the metadata is not available. This is in particular the case, if `computeNumberMatched: false`. This change ensures that `numberMatched` is always written; `unknown` is used, if the information is not available. `numberReturned` should always be available, but in case the information is not available `0` is written (`unknown` is not valid for this attribute).


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1659.surge.sh

PR #1663, opened 2026-06-24
Fix multi-collection responses

### LINKED ISSUE #1661: Search: incorrect encoding of features in multi-collection queries
In the Search building block, when executing a multi-collection feature response, the feature encoders apply the configuration of the arbitrary first collection to every feature. This results in incorrectly encoded features, if the encoding options of a format are collection-specific.

The configuration options for the feature format should always use the configuration of the collection the current feature belongs to.

### LINKED ISSUE #1662: Search: disable feature formats with fixed schemas
Currently also fixed-schema formats (such as FlatGeobuf or CSV) can be enabled for ad-hoc or stored queries in the Search building block. Since the queries are in general multi-collection queries where each collection / feature type will have a different schema, this cannot work.

These feature formats should be disabled in the Search building block.

From the current formats, only GeoJSON/JSON-FG, GML, HTML and the Debug Tokens should be enabled.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1661.
Closes #1662.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1663.surge.sh

azahnen (approved):

PR #1668, opened 2026-06-25
Transactions: configurable per-transaction setup and pre-commit hooks

cportele:
### Pull request checklist

-   [x] Updated documentation

A follow-up to #1627 and #1634.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/552.

Adds two API-level options to the TRANSACTIONS building block that run configurable SQL on the write transaction: `transactionSetup` (right after the transaction begins, e.g. `SET LOCAL ...`) and `preCommit` (right before commit, e.g. consistency checks or derived-data updates). Each option selects statements by the request's `Prefer: handling` preference through `always`/`strict`/`lenient` buckets. A pre-commit error rolls the transaction back and is reported; a warning lets it commit and is returned to the client.

- TransactionHooks: new value type with `always`/`strict`/`lenient` lists and effective(handling). TransactionsConfiguration gains `transactionSetup` and `preCommit`.
- TransactionExecutor.execute takes the `Prefer: handling` value (replacing the derived validate flag); executeAtomic and executeBatch run the setup and pre-commit statements and accumulate their warnings.
- ExecutionResult gains warnings; CommandHandlerTransactionsImpl renders a top-level warnings array, including on the no-content return path when warnings are present.
- Hook failures surface as FeatureMutationHookException carrying the warnings emitted before the failure; the executor recovers them and logs the configuration-driven rollback at debug without a stack trace.

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '7.0.0-SNAPSHOT'
 xtraplatform-native  = '2.6.0-SNAPSHOT'
-xtraplatform-spatial = '8.0.0-SNAPSHOT'
+xtraplatform-spatial = '8.0.0-transaction-lifecycle-hooks-SNAPSHOT'
 
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '7.0.0-SNAPSHOT'
 xtraplatform-native  = '2.6.0-SNAPSHOT'
-xtraplatform-spatial = '8.0.0-transaction-lifecycle-hooks-SNAPSHOT'
+xtraplatform-spatial = '8.0.0-SNAPSHOT'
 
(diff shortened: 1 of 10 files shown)


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1668.surge.sh

azahnen (dismissed):
My only comment would be that having SQL statements in the API configuration feels very wrong, especially since the module maturity is `CANDIDATE` and not `PROPOSAL`.
I guess it is fine for now, but there should be a clear statement in the `build.gradle` that this has to be resolved before a maturity promotion. If there is already an idea how that could look like, all the better.

cportele:
@azahnen - Agreed. Once there is experience about the type of hooks that are needed, we should update the configuration options. I will reduce the level to `PROPOSAL` for now and add the following in `build.gradle`:

> Transaction hooks in the configuration are currently SQL statements. Once there is a better understanding of the types of hooks/statements that are needed, the option should be changed to declaration that can be mapped to SQL statements internally. The maturity level should only be raised once this is implemented.

PR #1672, opened 2026-06-27
Search: single-shot and paging support; fix links

### LINKED ISSUE #906: Search: links in ad-hoc queries
A follow-up to #755: Properly handle links in responses to ad-hoc queries (links imply use of GET, but ad-hoc queries use POST with JSON payload). The currently links are not correct.

The `self` and `alternate` links should simply be suppressed, there is no resource on the server that we can link to (this is a transient ad-hoc query).

With respect to the `next` link, if there are more matches than the page size, here are some options:

- No support paging with ad-hoc queries. 
- Use an extended link object that includes the information to use the POST method and includes the request content and content type.  The content is the same as from the request - with an adjusted value for `offset`.
- Store that request for a few minutes and use a hash key to create an opaque next link.

The first option is not a good idea, paging should be supported for all requests that deliver features.

The third option also will not work, if the deployment consists of multiple instances, the next-queries would need to be stored in a shared store.

The second option is not ideal either, but probably the best approach and could be useful in other cases, too. The approach should also be discussed in OGC (maybe just remove the links for now and wait for that discussion). There is also a [somewhat related I-D](https://mnot.github.io/I-D/draft-nottingham-link-hint.html).

cportele:
### Pull request checklist

-   [x] Updated documentation

Follow-up to #1647.

Closes #906.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/555.

I had second thoughts on removing paging entirely from the Search endpoints. The reasons are: stored queries in current APIs use paging and paging support is at the moment still expected in OGC API Features Part 10. Also, for relatively simple stored queries paging can be useful, so it probably also should be kept in the future. 

However, often stored queries will be most useful as a single-shot, in particular, if they are complex, involve many sub-queries and return large numbers of features.  The Search endpoint responses (ad-hoc and stored queries) are now **single-shot by default**: all matching features are returned in one pass, without `numberMatched`/`numberReturned`. This suits long-running query expressions with high limits and many sub-queries - and asynchronous processing (a capability to be added in a follow-up PR). 

#1647 had dropped paging from Search endpoints; this PR makes single-shot the explicit default and **re-introduces paging as an opt-in** for stored queries, where the result is an addressable resource via a GET request.

For ad-hoc queries, which use POST there is no GET-addressable resource and paging is not supported. This resolves the main issue in #906. `self` and `alternate` links were already dropped for ad-hoc queries - except in HTML (see below).

#### Single-shot vs. paged

- **Single-shot (default).** One pass over the whole result set: no meta queries, no `numberMatched` / `numberReturned`, no paging links. `limit` does not page; an optional server-side maximum (`maximumFeaturesPerQuery` in the SEARCH building-block configuration) bounds the response. A server-side cursor keeps memory bounded regardless of result size.
- **Paged — `supportPaging: true`, stored queries only.** Reports `numberMatched` / `numberReturned`, is retrievable page by page via the `offset` query parameter, and includes `prev`/`next`/`first` links and the HTML pagination bar. For a multi-query, `numberMatched` is the invariant total across **all** sub-queries, while value queries are executed only for the sub-queries needed for the page.
- **Ad-hoc queries (POST) are always single-shot.** A POST request produces a transient result with no addressable resource to page through, so an explicit `supportPaging: true` is rejected with a clear error.

#### Links and HTML

An ad-hoc query is submitted via POST with a JSON query expression and produces a transient result with no addressable resource on the server. Its HTML representation nevertheless advertised a canonical (`self`) link and an offset-based pagination bar pointing at GET URLs the POST-only endpoint rejects.

`FeatureQueryScope` — carried by the feature transformation context and routed from the query handler into the HTML view — distinguishes `COLLECTION`, `AD_HOC_QUERY`, `STORED_QUERY` (paged) and `STORED_QUERY_SINGLE_SHOT`, and derives:

- `supportsPaging` — pagination and `prev`/`next` links are shown only for paged responses (collections, and stored queries with `supportPaging: true`). Query-expression pagination URLs use `offset` only (there is no `limit` query parameter for stored queries).
- `hasCanonicalResource` — the `canonical`/`self` link is suppressed only for ad-hoc queries; a stored query is a real GETable resource and keeps its `canonical` and `alternate`-format links.

The body and `Link` header were already correct (no links for ad-hoc); this aligns the HTML representation with the other formats.

## Renamed properties

- Query expression: `computeNumberMatched` → `supportPaging` (the query-level flag). The provider-level `computeNumberMatched` option is unchanged and still controls whether the count is computed at all.
- View flag: `fromStoredQuery` → `isQueryExpression` — it was set for ad-hoc queries as well and is used to drop the per-feature title link and keep the map data CRS, both of which apply to any query expression. The `feature.mustache` token is renamed to match. `isQueryExpression` on the transformation context now derives from the scope instead of the presence of a query id; the search handler is the only producer of query-id-bearing feature responses and sets the scope.

## Behaviour change / migration

- The default for query expressions changes from counted to **single-shot**. A stored query that relied on the previous default and wants paging must set `supportPaging: true`.
- `computeNumberMatched` is no longer a query-expression member; replace it with `supportPaging` (`computeNumberMatched: false` → `supportPaging: false`; both are single-shot). The counted, page-by-page behaviour is now opt-in via `supportPaging: true`.

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '7.0.0-SNAPSHOT'
 xtraplatform-native  = '2.6.0-SNAPSHOT'
-xtraplatform-spatial = '8.0.0-SNAPSHOT'
+xtraplatform-spatial = '8.0.0-multi-feature-query-scopes-SNAPSHOT'
 
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '7.0.0-SNAPSHOT'
 xtraplatform-native  = '2.6.0-SNAPSHOT'
-xtraplatform-spatial = '8.0.0-multi-feature-query-scopes-SNAPSHOT'
+xtraplatform-spatial = '8.0.0-SNAPSHOT'
 
--- a/ogcapi-draft/ogcapi-features-search/src/main/java/de/ii/ogcapi/features/search/app/EndpointAdHocQuery.java
+++ b/ogcapi-draft/ogcapi-features-search/src/main/java/de/ii/ogcapi/features/search/app/EndpointAdHocQuery.java
@@ -17,6 +17,7 @@ import de.ii.ogcapi.features.core.domain.EndpointRequiresFeatures;
 import de.ii.ogcapi.features.core.domain.FeatureFormatExtension;
 import de.ii.ogcapi.features.core.domain.FeaturesCoreConfiguration;
 import de.ii.ogcapi.features.core.domain.FeaturesCoreProviders;
+import de.ii.ogcapi.features.search.domain.ImmutableQueryExpression;
 import de.ii.ogcapi.features.search.domain.ImmutableQueryInputQuery;
 import de.ii.ogcapi.features.search.domain.QueryExpression;
 import de.ii.ogcapi.features.search.domain.SearchConfiguration;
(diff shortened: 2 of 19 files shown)


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1672.surge.sh

azahnen (approved):

PR #1676, opened 2026-06-29
improve performance of building-block identifier lookup

cportele:
### Pull request checklist

-   [ ] Updated documentation

ExtensionConfiguration.getBuildingBlockIdentifier() derives the identifier from the configuration class name via Guava CaseFormat on every call, and it is queried per feature on the extension-enablement path (ApiExtension.isExtensionEnabled). On large feature responses that case conversion showed up as a measurable share of CPU. Memoize the result per class (the identifier is a pure function of the class), removing the conversion from the hot path with no change in behaviour.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1676.surge.sh

azahnen (approved):

PR #1686, opened 2026-07-13
improve XML Schema validation

cportele:
### Pull request checklist

-   [ ] Updated documentation

Strict validation validates the document through a JAXP Validator with no hardening: FeaturesFormatGml.validate validated the WFS-T request body (reachable via Prefer: handling=strict) and StyleFormatSld10.analyze validated the stylesheet request body (on style create/replace with strict). The default Validator resolves DOCTYPE and external entities, so the body could read local files or reach internal URLs (XXE) or trigger entity expansion.

Add SecureXml in ogcapi-foundation as the single source for hardened XML processors and route every XML parse/validation through it:
- source()/harden(): validate an instance document through a reader that disallows DOCTYPE declarations and external entities (built fresh per call so it survives Validator.reset()), plus set ACCESS_EXTERNAL_DTD/SCHEMA to "" and enable secure processing on the validator.
- schemaFactory(): schema factory with secure processing (resolution of the admin-configured schemas still works).
- inputFactory(): hardened StAX factory.

Callers routed through SecureXml:
- FeaturesFormatGml (GML strict validation) and StyleFormatSld10 (SLD 1.0) — the previously unhardened validators; now fixed.
- WfsTransactionParser, WfsValueXmlConverter, VersionedMutationStrategy — the existing StAX parsers, previously hardened inline; VersionedMutationStrategy now also disables external entities (it set only SUPPORT_DTD=false before).

SecureXml lives in ogcapi-foundation because all callers are ldproxy building blocks; a class comment notes it should move to xtraplatform if a non-ldproxy class ever needs it. Adds SecureXmlSpec.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1686.surge.sh

azahnen (approved):

PR #1693, opened 2026-07-15
Positions in additional CRSs, associated profiles, ISO 19139 refinements

### LINKED ISSUE #1692: Geometries in additional coordinate reference systems
Normally, all geometries of a feature type are stored in a single CRS — the native CRS of the feature provider — and the API contract determines the CRS of a response: without a `crs` query parameter, geometries are returned in the default CRS (CRS84 or CRS84h); with `crs`, they are returned in the requested CRS.

Some datasets, however, record geometries in one of several CRSs and require that the geometry is preserved exactly as recorded. Four cases need special treatment:

- The geometry is in a CRS that is not the native CRS;
- CRSs that map to the same EPSG CRS (e.g. two geodetic realizations sharing one EPSG code, so the original identifier cannot be recovered from the CRS);
- coordinates that use a different convention than the CRS definition (e.g. German Gauss-Krüger coordinates written without the zone prefix, i.e. a false easting of 500,000 where the CRS defines 3,500,000); and 
- 1D vertical reference systems, which have no representation in a 2D/3D geometry model.

For such feature types, the provider should store each geometry in a property for its CRS, alongside the verbatim CRS URI. Where the CRS differs from the native CRS, a derived, transformed copy of the geometry is stored in the primary geometry property in the native CRS. The main geometry property declares the group in the provider schema:

```yaml
position:
  type: GEOMETRY
  variants:
    crsProperty: pos_srs          # the verbatim identifier, as recorded
    verticalProperty: pos_h       # 1D positions
    geometryProperties:
    - pos_gk3                     # 2D/3D positions, one property per CRS
    - pos_geoc
    - pos_h
pos_srs:
  type: STRING
  role: ORIGINAL_CRS_IDENTIFIER
  constraints:
    enum:
    - http://example.com/def/crs/DE_DHDN_3GK3
    - http://example.com/def/crs/ETRS89_X-Y-Z
    - http://example.com/def/crs/ETRS89_h
pos_gk3:
  sourcePath: position_gk3
  type: GEOMETRY
  role: ORIGINAL_GEOMETRY
  nativeCrs:
    code: 5677
  originalCrsIdentifiers:
  - urn:adv:crs:DE_DHDN_3GK3
  falseEastingDifference: 3000000
pos_geoc:
  type: GEOMETRY
  role: ORIGINAL_GEOMETRY
  nativeCrs:
    code: 4936
  originalCrsIdentifiers:
  - http://example.com/def/crs/ETRS89_X-Y-Z
pos_h:
  type: FLOAT
  role: ORIGINAL_HEIGHT
  unit: m
  originalCrsIdentifiers:
  - http://example.com/def/crs/ETRS89_h
```

The referenced properties are internal — they are read from the data source, but should not appear in queryables, sortables, schemas or as regular properties in responses. Each geometry variant declares which CRS URIs route to it and how its coordinates relate to the CRS it is stored in.

How the API responds is controlled by two profiles:

- **`crs-requested`** (the default) keeps the standard API contract: every feature encoding returns the primary geometry in the requested or default CRS. A feature whose recorded position cannot be represented in the requested CRS, such as a position in a 1D vertical reference system, is returned without a geometry (`"geometry": null` in GeoJSON, no position element in GML).

- **`crs-original`** returns the positions as recorded, unaffected by the `crs` query parameter, if supported by the featur format. In GML, the position element carries the verbatim identifier as `srsName` and the coordinates in the convention of that identifier; a 1D position is encoded as a point with `srsDimension="1"`. In JSON-FG, the recorded position appears in `place` with the identifier as `coordRefSys`, while `geometry` continues to hold the derived position in CRS84; for a 1D position, the vertical coordinate and its identifier appear in `properties`. Feature encodings that cannot represent positions in other reference systems, such as GeoJSON, ignore the profile. Ingesting documents written with `crs-original` reproduces the stored data exactly, so the profile supports lossless roundtrips.

A service whose clients expect the recorded positions by default can declare `crs-original` as its default profile for selected feature encodings (`defaultProfiles`), while clients can always select either behaviour explicitly via the `profile` query parameter.

### LINKED ISSUE #1694: CQL2: NOT and IS NULL filter predicates on joined properties never match
CQL2 predicates on properties that require a join in the SQL feature provider — feature references backed by a junction table, multi-valued properties, columns in joined tables — are encoded as an INNER-join semi-join:

`A.id IN (SELECT AA.id FROM <main> AA JOIN ... WHERE <column> <op>)`

This is correct for positive predicates ("there is a related row that matches"), but two constructs derived from it silently return no or wrong results instead of failing:

1. `NOT <predicate>` — the negation is pushed into the subquery by string surgery on the rendered SQL (last `IN` → `NOT IN`, last `WHERE` → `WHERE NOT`, …). That negates the inner predicate instead of the whole one, with two consequences:
  - Features without any related row contribute no subquery rows (INNER join) and are excluded from the negation, although they trivially satisfy it.
  - Multi-valued properties get exists-semantics ("some related value does not match") instead of the negation of the exists-semantics of the positive predicate ("no related value matches"). For example, `NOT role = 'x'` on a multi-valued role matches features that have the role 'x' alongside any other role, and does not match features without roles.
2. `<property> IS NULL` — `IS NULL` is applied to the joined column inside the subquery. Under the INNER joins the column is never null there, and features without related rows produce no subquery rows at all, so the predicate is unsatisfiable by construction. Filtering feature references with `rel IS NULL` ("has no reference") always returns an empty result set.

Since the outer operand of the semi-join (the sort key of the main table) is never null, negating the whole semi-join is exact and NULL-safe. The fix encodes `NOT <predicate>` as `NOT (A.id IN (SELECT …))` whenever the operand requires a join (the surgical rewrite is kept for plain columns, where it is equivalent), and `IS NULL` as `NOT (A.id IN (SELECT … WHERE <column> IS NOT NULL))`.

Note that this deliberately changes the result of negated predicates on multi-valued properties (from "has some non-matching value" to "has no matching value"), which is the consistent reading of logical negation in CQL2.

Additionally, an invalid filter coordinate (e.g. metric coordinates in a filter interpreted as CRS84 because filter-crs was not set) causes a 500 instead of a 400: a failed coordinate transform returns a CoordinateTuple whose `isNull()` is `true` rather than a `null` reference, but `CqlCoordinateChecker` only tests the reference and then fails with a `NullPointerException` when reading the coordinate values. The checker should test `isNull()` and report the invalid coordinate.

### LINKED ISSUE #1701: Search: offer HTML only for queries in the default CRS
The HTML representation only works for query results that use the default CRS (the embedded map does not support other CRSs) and that are paged (unpaged responses may be too large for a browser to render). HTML requests for other queries should be rejected with a 406 - and HTML should not be offered in the dropdown list in the HTML view.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1692. 
Closes #1694.
Closes #1701.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/577.

#### New building block PROFILE_CRS

For feature types with position variants, the profile set `crs` negotiates the representation:

- `crs-requested` (the default): the primary geometry is returned in the requested CRS (or the default CRS) in every feature encoding — the standard API contract. A feature whose position cannot be represented in the requested CRS (e.g. a 1D vertical position) is returned without a geometry.
- `crs-original`: positions are represented as recorded, identified by the stored verbatim identifier, unaffected by the `crs` query parameter. Supported by GML and JSON-FG; encodings that cannot represent such positions (e.g. GeoJSON) ignore the profile. A service can make `crs-original` its default via `defaultProfiles`.

#### GML

With `crs-original`, the original geometry, if provided, is encoded instead of the primary geometry: the verbatim identifier as `srsName`, the optional, inverse false-easting shift applied, 1D positions reconstructed with `srsDimension="1"`. Without it, the variant properties are suppressed and the primary geometry passes through. The `srsName` rendering for `srsNameStyle: TEMPLATE` and the decoding of non-OGC identifiers on ordinary geometries now use the `alternativeUri` declared on the `additionalCrs` entries of the CRS building block (e.g. `urn:adv:crs:ETRS89_UTM32` for EPSG:25832) — requests and `Content-Crs` continue to use OGC URIs.

Fix missing uom attribute on the members of a value array. The uom attribute (gml:MeasureType) was only written for properties of type FLOAT or INTEGER; for a VALUE_ARRAY the numeric type is the value type, so the members of a FLOAT/INTEGER array with a 'unit' in the provider schema were encoded without the attribute.

#### JSON-FG

With `crs-original`, a stored position is written to `place` with the verbatim identifier as `coordRefSys` inside the geometry object and no coordinate rounding; `geometry` keeps the derived position in CRS84. For a 1D position, the vertical coordinate and the identifier appear in `properties`.

#### ISO 19139 refinements (GML)

`valueWrap` chain entries support attributes and injected empty elements (`gmd:valueUnit[xlink:href=urn:adv:uom:m]/`, `gco:Record[xsi:type=gml:doubleList]`), enabling lossless roundtrips of ISO 19139 data-quality structures such as `DQ_QuantitativeResult`.

#### Search

The HTML representation only works for query results that use the default CRS (the embedded map does not support other CRSs) and that are paged (unpaged responses may be too large for a browser to render). HTML requests for other queries are now rejected with a 406 instead of silently using the default CRS. Responses of queries without paging are returned with 'Content-Disposition: attachment' so that browsers download the result instead of rendering it.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1693.surge.sh

azahnen (approved):

PR #1700, opened 2026-07-20
fix 500 response for an invalid value of a queryable query parameter

cportele:
### Pull request checklist

-   [ ] Updated documentation

- map `CqlParseException` and `NumberFormatException` raised while parsing the value of a queryable query parameter (e.g. an invalid instant or number) to `IllegalArgumentException` so that the API returns a 400 with a helpful message instead of a 500
- add a unit test for the parameter parsing


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1700.surge.sh

azahnen (approved):

PR #1705, opened 2026-07-23
transactions: option to collect all errors in atomic transactions

cportele:
### Pull request checklist

-   [x] Updated documentation

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/584.

Changes:

- new TRANSACTIONS option `collectErrors` (default `false`): a failing action no longer stops an atomic transaction — every action runs inside a database savepoint, all errors are collected and reported in the response, and all changes are still rolled back; falls back to stop-at-first-error when the provider session does not support savepoints
- new action status `ROLLED_BACK`: in a failed atomic transaction, actions that succeeded before the rollback no longer render as successes (the 422 body listed their feature URIs and counted them in summary)
- transaction-level errors (failed commit or pre-commit hook) are reported as a single exception instead of one duplicated exception per action
- summary gains `totalFailed`, `totalRolledBack` and `totalSkipped`
- exceptions and error texts use the canonical API collection id instead of the raw type name from the request
- per-action SQL warnings (RAISE WARNING / RAISE NOTICE from triggers or functions) are drained from the provider session and returned in warnings, prefixed with the action label
- specs for the atomic execution loop (savepoint lifecycle, error collection, fatal-error skipping, fallback) and the response rendering

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '7.0.0-SNAPSHOT'
 xtraplatform-native  = '2.6.0-SNAPSHOT'
-xtraplatform-spatial = '8.0.0-tx-savepoints-SNAPSHOT'
+xtraplatform-spatial = '8.0.0-SNAPSHOT'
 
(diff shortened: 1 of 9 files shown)


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1705.surge.sh

azahnen (approved):

PR #1706, opened 2026-07-26
gml: map properties to complete XML element chains

cportele:
### Pull request checklist

-   [x] Updated documentation

Adds support for additional differences between the provider schema and the GML representation, e.g., if a scalar feature property has to be mapped to an object value in GML.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/586.

`xmlPaths` maps a property path to the full chain of XML elements that represents the property, outer to inner: the first segment replaces the property element (a mapped property contributes neither its name nor its alias) and the innermost segment carries the value. Consecutive properties whose chains share a prefix share those elements, opened once and kept open, so a flat provider schema can be encoded as arbitrarily nested XML. This supersedes `valueWrap`, which could only wrap a value inside the property's own element.

- GmlConfiguration: `valueWrap` becomes `xmlPaths`, ValueWrapElement becomes XmlPathElement; documentation and examples updated. A chain segment keeps the attribute predicates and the trailing-`/` empty-element form, and `codeListUriTemplateIso19139` now keys off the chain's innermost element.
- GmlWriterXmlPaths opens, shares and closes the chain elements around a value; GmlWriterProperties delegates to it.
- The `uom` attribute goes on the innermost element, the one that carries the value, instead of on the outermost element of the chain.
- UpdatePathResolver resolves an input property path that addresses a property through its chain: the chain elements name no schema property, so they are consumed as a whole and resolve to the mapped property.

Bugfixes:

- MutationStrategy.extractPrimaryIntervalStart takes the chains and the body format's `useAlias` flag, so a Replace body's lifetime start is read from the element the payload actually carries. Without it the lookup missed a chain-mapped property and the retired predecessor's interval ended at the mutation time rather than at its successor's start.
- A retired feature answers 410 again instead of 404: the probe decides on the feature it returned rather than on `numberMatched`, which is absent whenever a provider runs with `computeNumberMatched: false`.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1706.surge.sh

azahnen (approved):

PR #1707, opened 2026-07-27
gml: encode object properties as XML element chains

cportele:
### Pull request checklist

-   [ ] Updated documentation

Follow-up to #1706. Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/587.

An xmlPaths chain may map an object property: the chain replaces both the property element and the object element, its innermost element holds the object's properties, and their own chains are relative to it. For an array of objects a leading '*' marks the segment from which the chain repeats per member: the segments before it wrap the array as a whole and are written once, the segments from it are written again for every member. Consecutive properties sharing leading segments continue to share those elements across an object chain.

- XmlPathElement parses the repetition marker; the configuration check rejects more than one marked segment per chain.
- XmlPathWriter opens, shares and closes the chain elements of values and objects; GmlWriterXmlPaths opens an object's chain at the object boundary and closes it at the object's end instead of closing the open wrappers; GmlWriterProperties contributes neither the property element nor the object element for a chained object.
- Objects with variable element names, feature references and the Link and Measure object types keep their own encoding and are never chained.
- The transactions path resolver ignores the marker when matching an input property path against a chain.

### DIFF
--- a/ogcapi-draft/ogcapi-transactions/src/main/java/de/ii/ogcapi/transactions/app/UpdatePathResolver.java
+++ b/ogcapi-draft/ogcapi-transactions/src/main/java/de/ii/ogcapi/transactions/app/UpdatePathResolver.java
@@ -230,9 +230,15 @@ final class UpdatePathResolver {
     return segment.trim().endsWith("/");
   }
 
-  /** The local name of a chain segment: its {@code prefix:} and attribute predicates removed. */
+  /**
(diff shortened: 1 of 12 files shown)


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1707.surge.sh

PR #1708, opened 2026-07-28
fix profile crs-original

cportele:
### Pull request checklist

-   [x] Updated documentation

The crs-original profile only replaced the positions of a geometry property with a crsVariants declaration; every other position was returned in the requested or in the default CRS of the API, which was not intuitive. The profile now also selects the CRS of the response, so that a position without a variant is returned in the CRS in which the feature provider stores it.

The CRS is a fallback for the default CRS of the API: a crs parameter in the request, or a crs in a query expression, takes precedence. In the HTML representation the profile is ignored, as is the crs parameter.

- ogcapi-features-core: new ProfileResponseCrs, applied in FeaturesCoreQueriesHandlerImpl before the CRS transformer is created - the profiles are negotiated first and the coordinate precision is recomputed for the selected CRS
- ogcapi-profile-crs: ProfileCrsOriginal returns the CRS of the feature provider; the profile set is no longer restricted to collections with a crsVariants declaration, which would make the positions of a query over multiple collections depend on the collections in the query
- ogcapi-features-search: the profiles are negotiated before the CRS of the query is determined; a profile that selects the CRS rules out HTML, both for a request (406) and for the formats offered for a stored query
- documentation, and specs for the CRS of the response and for the HTML rule

### DIFF
--- a/ogcapi-custom/ogcapi-profile-crs/build.gradle
+++ b/ogcapi-custom/ogcapi-profile-crs/build.gradle
@@ -1,11 +1,12 @@
 
 maturity = 'PROPOSAL'
 maintenance = 'NONE'
-description = 'Profile for positions stored in a reference system that differs from the storage CRS.'
-descriptionDe = 'Profil für Positionen, die in einem anderen Referenzsystem als dem Speicher-CRS gespeichert sind.'
+description = 'Profiles for the reference system of the positions in the response.'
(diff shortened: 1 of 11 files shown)


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1708.surge.sh

PR #1712, opened 2026-08-02
gml: version-unique gml:ids via the versions-as-features-unique-ids profile

cportele:
### Pull request checklist

-   [x] Updated documentation

A response that contains multiple versions of a feature requires unique gml:id values. This was previously handled by two overlapping mechanisms: the versions-as-features-unique-ids profile (rewriting the feature id to a composite of the canonical id and the version's interval start) and the GML-only option appendTemporalSuffixToGmlId (suffixing only the gml:id attribute). The profile, which GML already declares as its default for the versions profile set on versioned collections, is now the single mechanism and the unreleased option is removed.

The profile so far only worked on /items: query expressions (ad-hoc and stored queries) negotiated profiles but never applied their query transformations, so e.g. a stored query with f=xml returned colliding gml:id values.

- ProfileFeatureQuery: transformMultiFeatureQuery hook for query expressions, applied by SearchQueriesHandlerImpl to the multi-query
- ProfileVersionsAsFeaturesUniqueIds: implement the hook; no interval gate, a query expression applies no snapshot default and can always return multiple versions of a feature
- remove appendTemporalSuffixToGmlId and the gml:id suffix machinery (option, docs, CollectionEncodingGml flag, suffix state and writers)
- GmlWriterId keeps the canonical feature id on gml:identifier when the id was rewritten
- document the versions entry of the built-in GML defaultProfiles
- specs: multi-query transformation, gml:identifier override


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1712.surge.sh

PR #1725, opened 2026-08-07
 Seeding the source tilesets from the tiles provider

### LINKED ISSUE #1722: Source tilesets are not seeded when they have no collection
### ldproxy Version

7.8.0

### Current Behavior

With one tileset per collection, seeding works as expected. The problems below appear once the tilesets of a tile provider no longer map onto the collections of the API — which happens as soon as a combined tileset is built from sources that are not published themselves, such as label points or derived symbol positions.

Such tilesets without a collection are never seeded. TileSeedingBackgroundTask.getJobSet builds the job set from the collections plus the dataset tileset, so a tileset that exists only as a layer of a combined tileset gets no job. TileEncoderMvt.combine then amplifies this: it requests every source tileset through the full provider chain, and a cache miss falls through to the generator. For an unseeded source, combining costs a feature query per layer per tile, so the combine phase can cost more than generating all tilesets did — after seeding appeared to succeed.

In addition:
 
- Seeded tiles and empty tiles are decided by different extents. What gets seeded for a collection tileset follows the extent of the collection; sparse and the generator use the extent of the tileset's feature type. Identical in the normal case, but they diverge when the tile provider reads from a different feature provider than the features of the API, or when the extents disagree at the margin. Tiles inside the data extent are then never seeded and get regenerated on every combine.
- The job set announces tilesets that are not seeded. A tileset whose zoom levels miss the levels of every seeded cache has nothing to seed but is still listed as scheduled — misleading during a long run.

### Expected Behavior

All non-empty tilesets of a tile provider should be processed when tiles are seeded.

Approach:

TileSeedingBackgroundTask (ldproxy): the job set becomes the closure over combine — start from the tilesets the API publishes, then add those they are combined from, expanding *. The tiles to seed come from TilesetMetadata.getBounds() with the collection extent as fallback, so a single extent governs what is seeded, when sparse treats a tile as empty, and whether a layer contributes to a combined tile. Tilesets no seeded cache covers are dropped from the job set.

TileEncoderMvt (xtraplatform-spatial): before asking the provider chain for a layer, test the tile against the source tileset's bounds and skip the layer when they do not intersect — no lookup, no generation, no empty tile written. The tile bounds are transformed once per combined tile, and an unknown extent counts as "may have data" so nothing is skipped when metadata is missing. Deliberately not gated on sparse: an extent is a property of the data, whereas sparse only decides whether the generator checks before querying.

### Steps to Reproduce

Add tilesets to a tile provider that are not mapped to a collection.

### Relevant log output

```shell

```

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1722.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/596.

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '7.1.0-SNAPSHOT'
 xtraplatform-native  = '2.7.0-SNAPSHOT'
-xtraplatform-spatial = '8.1.0-tiles-combine-skip-empty-layers-SNAPSHOT'
+xtraplatform-spatial = '8.1.0-SNAPSHOT'
 
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '7.1.0-SNAPSHOT'
 xtraplatform-native  = '2.7.0-SNAPSHOT'
-xtraplatform-spatial = '8.1.0-SNAPSHOT'
+xtraplatform-spatial = '8.1.0-tiles-combine-skip-empty-layers-SNAPSHOT'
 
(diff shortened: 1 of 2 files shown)


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1725.surge.sh

azahnen (approved):

PR #1728, opened 2026-08-08
fix XML attributes emitted before the GML feature element

### LINKED ISSUE #1726: GML: XML attributes emitted before the GML feature element
### ldproxy Version

4.8.0

### Current Behavior

A property mapped to an XML attribute was rendered in front of the element it belongs to: before the feature element for a single-feature request, and before the member wrapper for a feature collection.

This is a regression that likely was introduced in the GML changes in v4.8.0.

### Expected Behavior

XML attributes are written correctly.

### Steps to Reproduce

Declare a feature property via `xmlAttributes`.

### Relevant log output

```shell

```

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1726.

A property mapped to an XML attribute was rendered in front of the element it belongs to: before the feature element for a single-feature request, and before the member wrapper for a feature collection.

The attribute is injected through a placeholder that is appended straight to the encoder's output buffer, so that it lands in attribute position of a start tag whose '>' is still pending. The XMLStreamWriter the runtime resolves buffers its output internally, so the append landed in front of everything the writer had not emitted yet. write() now flushes the writer first, which pins the raw string at the current position and leaves an open start tag open.

The same defect affected the gml:surfaceMember placeholder of CityGML LoD2 solids, which is written the same way.

- FeatureTransformationContextGml: flush the writer in write(); it and writeXmlAttPlaceholder() now declare IOException
- XmlAttributePlaceholderSpec: assert the emitted bytes of the real encoder for an item and a collection response, and that an unfilled placeholder leaves no trace
- ogcapi-features-gml: run the specs against the StAX implementation the runtime provides, and add the dependencies a stubbed OgcApi and ApiRequestContext need
- ForceCompositeCurveSpec: use Spock block labels, so its features run


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1728.surge.sh

azahnen (dismissed):
@cportele 
I think the explicit dependency on `woodstox` is unnecessary and error prone. Every module and also its tests already depend on `xtraplatform-runtime-tpl`, which contains `woodstox`. Only depending on the runtime means any changes to the XML stack there would  be automatically applied here as well, i.e. we would always test the actual XML stack that is also used at runtime.
The dependency may be verified by running `./gradlew dependencies --configuration testRuntimeClasspath` inside of `ogcapi-stable/ogcapi-features-gml`.

I removed it locally and the tests still succeed. If you agree, remove the dependency and merge.

PR #1729, opened 2026-08-08
gml: option to encode properties as annotation comments

### LINKED ISSUE #1727: GML: add option to include a scalar property as a comment

Add an option to include a scalar property as a comment in GML.

### Describe the Use Case

Such a capability is needed, if the property is not included in the GML application schema, but when the information is needed downstream to properly process the data.

Add a configuration option `xmlComments` following the pattern introduced with `xmlAttributes`.

### Additional Information

_No response_

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1727.

Depends on #1728.

Values that have no place in the GML application schema could so far only be conveyed as an XML attribute or an element, either of which makes the response invalid against that schema. A property listed in the new xmlComments option is encoded as an XML comment instead, which is allowed anywhere in element content, so the response stays valid. The property links of a feature already use this form; the escaping is now shared.

The comment is written in the position of the property element rather than collected at the start of the feature: the value keeps its order among its siblings, and a property of a nested object stays inside that object. A property listed as both an XML attribute and a comment is encoded as an attribute.

A comment is not part of the XML information a request body is decoded from, so such a property could not be sent back. Only properties excluded from the RECEIVABLE scope may therefore be listed, which keeps encoding and decoding symmetric by construction and leaves the decoder unchanged. Listing a receivable property, an unknown property, or configuring the option without a provider is reported at startup.

- GmlConfiguration: the option, its documentation and the merge rule
- CollectionEncodingGml, FeaturesFormatGml: the alias-rewritten property paths
- GmlWriterProperties: encode a listed property as an annotation comment
- FeatureTransformationContextGml: writeAnnotationComment() and the comment-text escaping, which GmlWriterPropertyLinks now shares
- GmlBuildingBlock: validate the option at startup; the building block gains its first dependency, so the configuration registry passes it on
- ldproxy-cfg: the option in the entity schema and the reflection configuration, without adopting the unrelated drift a full regeneration would bring
- AbstractGmlOutputSpec: the harness for specs that assert emitted bytes, pulled out of XmlAttributePlaceholderSpec
- XmlCommentsSpec, XmlCommentsValidationSpec: encoding, position, escaping, precedence and the scope restriction


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1729.surge.sh

azahnen (changes requested):
@cportele 
This also has the unnecessary `woodstox` dependency and some conflicts with master that need to be resolved. Otherwise it looks good and can be merged.

PR #1731, opened 2026-08-08
OGC API Processes

### LINKED ISSUE #1730: OGC API Processes

Add support for OGC API - Processes - Part 1: Core (v2.0).

### Describe the Use Case

Describe and execute processes.

_No response_

### Additional Information

_No response_

neiroukh:
### Pull request checklist

-   [x] Updated documentation

Closes #1730 .


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1731.surge.sh

neiroukh:
@cportele - two questions where I'd value your judgment:

1. [**Requirement 2, B**](https://docs.ogc.org/DRAFTS/18-062r3.html#req_core_landingpage-success): The conformance link uses HTTPS (`https://www.opengis.net/def/rel/ogc/1.0/conformance`), but we currently use HTTP. Should I update the link?

2. [**Requirement 16, B and C**](https://docs.ogc.org/DRAFTS/18-062r3.html#req_core_process-exception-no-such-process): This requires a specific exception type when accessing a non-existing process. In our case, the framework handles this automatically since the `processId` path parameter fails schema validation before it gets that far. Can I keep it this way or is it important to match the specific exception type?

neiroukh:
Please note that the permissions to dismiss jobs and execute processes are currently set to `READ`, see `JobQueriesHandler` and `ExecutionQueriesHandler`. This makes developing easier. Should be changed before merging.

cportele:
@neiroukh 

- [Requirement 2, B](https://docs.ogc.org/DRAFTS/18-062r3.html#req_core_landingpage-success): Since the requirement uses a HTTPS URI, we have to use that one.
- [Requirement 16, B and C](https://docs.ogc.org/DRAFTS/18-062r3.html#req_core_process-exception-no-such-process): Since it is a requirement, we have to conform to that, otherwise conformance tests will fail. 
  - Note that we already support RFC 9457 (successor of RFC 7807) for all JSON exception responses, but not in the way OGC API Processes requires it. We will have to upgrade the current exception mechanism first: 
    - extend `ApiErrorMessage` with an option for an explicit type (not the default "about:blank");
    - extend `ExceptionFormatJson` and `OgcApiErrorView`/`exception.mustache` with the type information;
  - Then we probably need a mechanism to handle such special cases, maybe in `ExceptionMapper`.

cportele:
@neiroukh - On the conformance link relation URI, I changed my mind. Leave that for now at HTTP. OGC API Common also uses HTTP, so I guess this must be an error that will be fixed before publication.

neiroukh:
@cportele - thanks for your valuable input! Then let's proceed this way.

neiroukh:
> @neiroukh
> 
>     * [Requirement 2, B](https://docs.ogc.org/DRAFTS/18-062r3.html#req_core_landingpage-success): Since the requirement uses a HTTPS URI, we have to use that one.
> 
>     * [Requirement 16, B and C](https://docs.ogc.org/DRAFTS/18-062r3.html#req_core_process-exception-no-such-process): Since it is a requirement, we have to conform to that, otherwise conformance tests will fail.
>       
>       * Note that we already support RFC 9457 (successor of RFC 7807) for all JSON exception responses, but not in the way OGC API Processes requires it. We will have to upgrade the current exception mechanism first:
>         
>         * extend `ApiErrorMessage` with an option for an explicit type (not the default "about:blank");
>         * extend `ExceptionFormatJson` and `OgcApiErrorView`/`exception.mustache` with the type information;
>       * Then we probably need a mechanism to handle such special cases, maybe in `ExceptionMapper`.

Update: I extended `ApiErrorMessage`, `ExceptionFormatJson`, `OgcApiErrorView` and `exception.mustache` to support the `type` property.

In `ExceptionMapper` i added the following if-statement to `processException(WebApplicationException exception, ExceptionFormatExtension exceptionFormat, String msg)`:

```java
// ToDo: Harden and improve type handling
if (msg.contains("processId")) {
  return Response.fromResponse(response)
      .type(exceptionFormat.getMediaType().type())
      .entity(
          exceptionFormat.getExceptionEntity(
              new ApiErrorMessage(
                  response.getStatus(),
                  response.getStatusInfo().getReasonPhrase(),
                  msg,
                  null,
                  "https://www.opengis.net/def/exceptions/ogcapi-processes-1/1.0/no-such-process")))
      .build();
}
```

While this simple mechanism works, it's probably a very bad idea due to its broad nature. Therefore I am still looking for a better way to check if the exception needs a type.

PR #1734, opened 2026-08-08
search: exclude properties from a query expression

### LINKED ISSUE #1733: Search: Support 'excludeProperties' in Queries

Currently only a `properties` projection can be specified, a projection to only exclude specific properties does not exist. This is a difference from the query parameters in the Projections building block.

### Describe the Use Case

Same as in the Projections building block.

Support a new member `excludeProperties`.

### Additional Information

_No response_

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1733.

A query expression could only select the properties to include, with "properties". A client that wants everything except a few had to enumerate the rest, and a stored query written for one downstream purpose could not drop a property that only another purpose needs.

Add "excludeProperties" alongside "properties", available both globally on the query expression and locally on a single query, and in both the stored form (where it accepts parameters like its counterpart) and the resolved form.

The effective field list of a sub-query: the global and local selection lists are concatenated, or all properties of the collection if neither selects anything; the global and local exclusion lists are unioned and subtracted from that. A global exclusion is applied to every queried collection, so a name a collection does not have is a no-op there; a name that none of the queried collections has is rejected, as is a name a local exclusion gets wrong, so that a typo does not pass silently. Excluding every property is rejected too -- an empty field list would otherwise be read as "all properties" downstream.

The resolution lives in PropertySelection so it can be tested on its own; PropertySelectionSpec covers the table and a round-trip of the new member at both levels.

Resolving an exclusion needs all property names of a collection, which is a full pass over its schema and is not cached upstream. A query expression can have hundreds of sub-queries over dozens of collections, so the names are cached per API version and collection, as the query parameters do for their schemas.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1734.surge.sh

azahnen (approved):

PR #1736, opened 2026-08-09
fix intermittent 500 responses for the first requests after startup

### LINKED ISSUE #1735: intermittent 500 responses for the first requests after startup
### ldproxy Version

4.8.0

### Current Behavior

Lazily populated parameter/schema caches keyed by the API hash used a check-then-act idiom on nested ConcurrentHashMaps. Under concurrent first requests, a second thread could replace an inner map that another
thread had just populated; the losing request then reads a null schema and fails with a NullPointerException in ParameterExtension.validateSchema.

### Expected Behavior

All requests complete without NPE.

### Steps to Reproduce

Start an API and send many parallel requests of the same kind.

### Relevant log output

```shell

```

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1735.

Lazily populated parameter/schema caches keyed by the API hash used a check-then-act idiom on nested ConcurrentHashMaps. Under concurrent first requests, a second thread could replace an inner map that another thread had just populated; the losing request then read a null schema and failed with a NullPointerException in
ParameterExtension.validateSchema.

- convert all lazily built caches to ConcurrentHashMap.computeIfAbsent: the first caller computes the value, concurrent callers for the same key wait until it is installed, so a null is no longer observable (15 parameter/header classes plus FeatureSchemaCache, JsonSchemaCache, SchemaGeneratorFeatureOpenApi and Metadata3dSchemaCacheImpl, which drops its synchronized double-check)
- guard ParameterExtension.validateSchema against a null schema: log a warning and skip schema validation instead of failing the request; parameter-specific checks in validateOther still run
- add ParameterExtensionSpec locking the null-schema guard


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1736.surge.sh

azahnen (dismissed):
Looks good, conflicts with master need to be resolved.

PR #1738, opened 2026-08-10
fix(html): filter delete button doesn't remove applied filters

### LINKED ISSUE #1737: Filter cannot be deleted in HTML filter editor — old value stays applied and blocks selecting a different one
### ldproxy Version

4.7.2

### Current Behavior

Filters cannot be deleted. For example, if I want to search for different municipalities one after another, I first enter "Gemeinde = Viersen". When I then click "Add", the condition restricting the search to the municipality of Viersen appears below the value selection field. Below this block I now have the option to either confirm or delete the selection. When I delete my selection, the selection block disappears, but the filter itself does not. I cannot select a different municipality.

### Expected Behavior

This means that the filter should actually be deleted once I click the "Delete" button, and that I should then be able to create a different expression for the same attribute.

### Steps to Reproduce

1. Open https://gdi-api.krzn.de/kvie_denkmale/collections/baudenkmal/items?f=html
2. In the filter dialog, select the field Gemeinde (municipality) = Viersen and click "Add"
3. Remove this filter setting again using the "Delete" button
4. Select a different municipality

### Relevant log output

```shell

```

PZahnen:
### Pull request checklist

-   [ ] Updated documentation

Closes #1737 .


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1738.surge.sh

azahnen (approved):

PR #1739, opened 2026-08-10
fix stack traces in the log for transactions the database rejects

cportele:
### Pull request checklist

-   [ ] Updated documentation

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/600

A transaction action rejected by a database integrity constraint — a CHECK or foreign-key constraint, a unique index, or a trigger raising one — is caused by the data the client sent, and the rejection is already reported in the response. It was logged at WARN with a full stack trace anyway, because the arriving exception is an IllegalStateException wrapping the driver exception and isUserError() only recognised IllegalArgumentException.

- isUserError() walks the cause chain and accepts the new FeatureMutationConstraintException, so these rejections log as a one-line WARN like a malformed payload does; genuine failures keep their stack trace
- also quietens duplicate-key and other client-caused rejections, which had the same problem


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1739.surge.sh

azahnen (approved):

PR #1740, opened 2026-08-11
transactions: fix oversized response for Prefer return=none

cportele:
### Pull request checklist

-   [ ] Updated documentation

A client asking for `return=none` gets 204 No Content only when the transaction succeeded and produced no warnings. In every other case the response fell through to the full representation, so a request that asked for nothing received the largest body available: one URI per written feature, which on a batch transaction can be many entries.

`return=none` now renders the same minimal body as `return=minimal` when a response is unavoidable: semantic, summary, exceptions and warnings, without the per-feature URI arrays.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1740.surge.sh

PR #1744, opened 2026-08-15
fix: sortby accepted any property when no sortables are configured

cportele:
### Pull request checklist

-   [ ] Updated documentation

If sorting was enabled for a collection, but the list of sortables was empty, the sortby parameter accepted any property of the feature type, because the enum of allowed sort keys was only set for a non-empty list of sortables.

- the sortby parameter is no longer available for collections that have neither sortables nor a default sort order
- if only a default sort order is configured, the parameter remains available, but all client-provided sort keys are rejected
- add a spec for the sortby parameter


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1744.surge.sh

PR #1745, opened 2026-08-18
gml: cache a compiled XSD schema per configuration instead of per API

### LINKED ISSUE #1749: GML: Improve XML Schema cache
The cache was keyed by the API's data, so a service reload produced a different key: every reload parsed all transitive XSDs of the catalog again and kept the previous grammar pool reachable for the lifetime of the process. Since a schema is derived from the schemaLocations and the catalog that resolves their imports, it should be keyed by those instead. An unchanged configuration would reuse its schema across reloads and across APIs.

cportele:
### Pull request checklist

-   [ ] Updated documentation
-   [ ] All checks are passing

Closes #1749.

The cache was keyed by the API's data, so a service reload produced a different key: every reload parsed all transitive XSDs of the catalog again and kept the previous grammar pool reachable for the lifetime of the process. Since a schema is derived from the schemaLocations and the catalog that resolves their imports, it is now keyed by those instead. An unchanged configuration therefore reuses its schema across reloads and across APIs, the cache is bounded by the number of distinct configurations, and a reload no longer pays for building a schema at all.

The catalog belongs in the key because it decides which files the imports resolve to; it enters as a hash, since it holds one entry per mirrored XSD and the key is built on every validation. The schema is built outside the map, because newSchema() is far too slow to run while holding a bin lock.

The two remaining caches keyed by the API's data hold per-collection values that are cheap to rebuild, so they are dropped when the API shuts down — which is also what a reload does before starting the next generation.

SchemaCacheKeySpec locks the properties the bound relies on: insensitive to the order of the schemaLocations, sensitive to a changed location and to a changed catalog, and stable for a configuration without schemaLocations.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1745.surge.sh

azahnen (approved):

azahnen (changes requested):
This does not compile anymore after merging the current master. #1757 added `canValidate`, which uses the deleted `schemaLocationsFingerprint`.

PR #1757, opened 2026-08-23
CRUD: fix defects and conformance gaps

### LINKED ISSUE #1756: OGC API - Features - Part 4 (Create, Replace, Update, Delete): defects and conformance gaps
Running the executable Abstract Test Suite (Annex A) of the draft **OGC API - Features -
Part 4 / OGC API - Common - Part 5: Create, Replace, Update and Delete** (OGC 20-002r2)
against ldproxy found 7 defects and 10 smaller deviations from the draft. Most are in the
`CRUD` building block, several also affect `TRANSACTIONS`, and four are in the feature
decoders in `xtraplatform-spatial`.

## Conditional requests / optimistic locking

- **`If-Unmodified-Since` was not evaluated — silent lost updates and deletes.** With
  `optimisticLockingLastModified: true` the header is required, but its value was ignored: a
  request with an outdated timestamp modified or deleted the feature instead of failing with
  `412`. The last modification time was read only from the representation used in mutation
  requests, and a server-managed timestamp is normally excluded from the `RECEIVABLE` scope,
  so it was never selected. It is now taken from the returnable representation when the
  mutation representation does not carry it, and a collection that enables the option
  although the provider type has no `isLastModified` property is reported at startup.
- **The timestamp was parsed with `Instant.parse` inside a `catch (Throwable ignore)`.** That
  fails for every value without a time zone — which is what the SQL provider delivers — so in
  the streams where the `DATE_FORMAT` transformation does not run, optimistic locking was
  disabled without a trace in the log. Now parsed with the same flexible parser as the
  temporal extent, interpreting a zoneless value in the provider's `nativeTimeZone`, and an
  unparseable value is logged instead of ignored. *(xtraplatform-spatial)*
- **`If-Match` was never evaluated.** No entity tag of a feature is known in a request that
  changes it, so the precondition cannot be met (RFC 9110, 13.1.1); the request is now
  answered with `412` instead of being executed with the header ignored.
- **A successful mutation did not report `Last-Modified`,** so a client could not obtain the
  value for the precondition of its next request.
- **A missing mandatory precondition was answered with `400`** instead of `428`
  (RFC 6585).
- **A precondition was required for a create, and evaluated before the existence check.** A
  PUT that creates a feature was rejected unless it stated a precondition, although there is
  no previous state to protect, and a request for a feature that does not exist was answered
  with a precondition status code instead of `404`. RFC 9110, 13.2.1, requires preconditions
  to be ignored when the response without them would be neither 2xx nor `412`. The check moved
  to where the current feature has already been fetched, and the header is required only if
  the feature has a modification date (RFC 9110, 13.1.4).

## Request bodies

- **A mutation request without a `Content-Type` header failed with `500`.** Now `400`.
- **`Prefer: handling=strict` did not enforce required properties.** The generated
  GeoJSON/JSON-FG validation schema dropped the required properties of the logical schema.
  Whether a request body can be validated is now a property of the format; without a schema
  the body is not validated, instead of being validated against the logical schema, which
  rejected every valid feature. A rejection reports the applied preference.
- **A JSON-FG request body with an in-document `coordRefSys` was rejected** with
  `Expected STRING, OBJECT or ARRAY for 'coordRefSys', but got: OBJECT` — the member was
  passed to the parser as the enclosing geometry object. *(xtraplatform-spatial)*
- **The coordinates of a request body were not restricted to the coordinate reference systems
  of the collection** — neither in the `Content-Crs` header nor in the body (a JSON-FG
  `coordRefSys` member, a GML `srsName` attribute). The JSON and GML geometry decoders can now
  be restricted to a list of coordinate reference systems, comparing the code only, since the
  axis order is a property of the encoding. *(decoders in xtraplatform-spatial)*
- **`Content-Crs` was honored on an API without CRS support.** Such an API does not declare
  "Coordinate Reference Systems by Reference", so per `/req/features/crs-crs84` the
  coordinates are always CRS84 and any other declared CRS has to be rejected; instead the
  native CRS of the provider was accepted and the coordinates transformed as if given in it.
  The building-block check was applied in `getSupportedCrsList` only for the API, not for a
  collection, while every request-side check goes through the latter.
- **A JSON Merge Patch that changes the `id` property was not rejected.** Now `400`; an `id`
  member carrying the identifier of the feature is still accepted.

## Other behaviour

- **DELETE ignored the `featureType` mapping of `FEATURES_CORE`,** and a feature change was
  reported for the collection instead of for the type in the provider, which clears its caches
  by that name.
- **OPTIONS responses were content-negotiated,** so a request with an `Accept` header that does
  not allow `text/plain` got `406`. An OPTIONS response is about the communication options for
  the resource, not a representation of it (RFC 9110, 9.3.7).
- **`Preference-Applied` claimed effective defaults as applied.** It now lists exactly the
  preferences that were submitted and honored, and is omitted when none were sent.
- **A source schema validator warning** was added for a `DATETIME` property mapped to a column
  with a time zone while `nativeTimeZone` is set to a zone other than UTC — such values are
  read in UTC and would then be reinterpreted, which shifts them. *(xtraplatform-spatial)*

## API definition and discovery

- **The conformance class URIs were the `0.0` draft URIs.** Now the URIs of the current draft,
  co-branded as OGC API - Common - Part 5 except for the class "Features".
- **The "Handling Preference" class was not declared** for an API whose formats can validate
  request bodies.
- **The API definition did not describe the responses of the CRUD operations** — no `201` for a
  PUT that creates a feature (with its `Location` header), and no `412` / `428` for a
  collection with optimistic locking, nor the `If-Unmodified-Since` parameter those depend on.
- **OPTIONS responses did not advertise the media types accepted for a request body.** They now
  include `Accept-Post` and `Accept-Patch` (W3C LDP 1.0, 4.5.2; RFC 5789, 3.1), derived from
  the request body of the POST and PATCH operation of the resource.

## CORS (follow-up to #1746)

- **`Access-Control-Allow-Credentials: true` was sent next to `Access-Control-Allow-Origin: *`.**
  A browser rejects a credentialed cross-origin response that allows any origin, so the pair
  granted nothing that works, while pre-arming the classic misconfiguration — this API accepts a
  cookie as a credential. Removed from both the response filter and the OPTIONS endpoint.
- **`Access-Control-Expose-Headers` named `Prefer`,** a request header that never appears in a
  response, instead of `Preference-Applied`; and it reported `Location` for POST only, although
  a PUT returns it as well when the request created the feature.

## Also affects the `TRANSACTIONS` building block

The `coordRefSys` fix, the validation-schema and "no schema, no validation" behaviour of
`handling=strict`, the OPTIONS content negotiation, the provider type in the change
notification, and the coordinate reference system of geometries in a request body are not in
`ogcapi-crud` and therefore apply to `/transactions` as well.

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1756

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/617.


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1757.surge.sh

PR #1765, opened 2026-08-28
fix rejection of queryable query parameter values with a wildcard

### LINKED ISSUE #1764: Queryables as query parameter validation too strict
### ldproxy Version

4.8.0

### Current Behavior

Queryables of a collection are by default (option `asQueryParameters`) mapped to query parameters of the `items` endpoint. Query parameters are filtered for equality. For string valued properties, ldproxy also supports `*` as a wildcard. 

For properties with constraints (e.g., a `regex` constraint) the use of a `*` wildcard may result in a value that validates the schema of the property in the provider schema. For example, a property with a regex `^[A-Z]{2}[0-9]{5}$` would reject a query parameter with a value "AB1*". 

### Expected Behavior

If a wildcard is used in the value, the constraints should no longer be included in the validation of the query parameter value.

### Steps to Reproduce

Define a regex on a queryable property. Use a value that includes "*" and violates the pattern.

### Relevant log output

```shell

```

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1764.

A queryable query parameter was always validated against the schema of the property, so a value with a wildcard was rejected whenever the property has constraints that the wildcard pattern cannot satisfy, for example a regular expression, an enum or a length restriction.

The values are available when a parameter is validated, only the typed parsing happens later, so the schema can be selected per request. Add the hook getSchemaForValidation() to ParameterExtension and use it in validateSchema(). The schema that is published in the API definition is unchanged.

For a queryable that supports a wildcard, that is a string value or an array of strings, a value with a wildcard is now validated against a plain string schema. A string schema also for arrays, because such a value is a single pattern that is matched against the array members and must not be split at the comma. Values without a wildcard, and queryables that do not support one, keep the schema of the property and are still rejected when they violate it.

### DIFF
--- a/ogcapi-stable/ogcapi-collections-queryables/src/main/java/de/ii/ogcapi/collections/queryables/app/QueryParametersQueryables.java
+++ b/ogcapi-stable/ogcapi-collections-queryables/src/main/java/de/ii/ogcapi/collections/queryables/app/QueryParametersQueryables.java
@@ -145,6 +145,13 @@ public class QueryParametersQueryables
     } else {
       description.append('.');
     }
+    if (QueryParameterTemplateQueryable.supportsWildcard(
+        queryableDefinition.getValue().getType(), queryableDefinition.getValue().getValueType())) {
+      description.append(
+          " The value may contain '*' as a wildcard for any sequence of characters. Schema"
+              + " constraints of the property, for example 'pattern' or 'enum', are not applied to"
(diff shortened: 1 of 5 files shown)


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1765.surge.sh

azahnen (approved):

PR #1768, opened 2026-09-01
CRUD/TRANSACTIONS: add an option to reject empty values

### LINKED ISSUE #1767: Transactions/CRUD: option to reject empty string values
With `handling=strict` the feature should not only be validated against the schema, but there should be an option to reject empty/blank string values. These values may be strictly schema-valid, but semantically this is in almost all cases not an acceptable value.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1767.

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/628.

`rejectEmptyValues` refuses a value that is a string without characters or with only whitespace. Like the schema validation it joins, the check only applies to a request with a `Prefer` header of "handling=strict"; unlike it, the check runs while the request body is decoded, so it needs no schema and also applies where none is available for the collection. A property the body omits, or states as null, is left without a value and is unaffected.

For a transaction it covers the payload of every insert and replace action as well as the values an update action sets. An update carries the values instead of a feature payload, so it never passes a decoder and is checked on the resolved values, reported the same way. Because the check rides the decoding rather than running before the write, an empty value fails the whole action, also under batch semantics, rather than skipping the item that carries it.

PATCH is not covered: it has no handling preference to opt in with.

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '7.1.0-SNAPSHOT'
 xtraplatform-native  = '2.7.0-SNAPSHOT'
-xtraplatform-spatial = '8.1.0-SNAPSHOT'
+xtraplatform-spatial = '8.1.0-reject-empty-values-SNAPSHOT'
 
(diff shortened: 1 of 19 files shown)


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1768.surge.sh

azahnen (dismissed):

PR #1770, opened 2026-09-02
MCP: fix HTTP 500 on every request after schema validator upgrade

### LINKED ISSUE #1769: MCP Server fails in v4.8
Intializing the MCP Server results in a 500 response in ldproxy 4.8. The reason is that the MCP code expects the JSON Schema validator in version 1.5, but ldproxy has upgraded to version 2.0.

cportele:
### Pull request checklist

-   [ ] Updated documentation

Closes #1769

The MCP SDK ships a default JsonSchemaValidator compiled against json-schema-validator 1.5.x. Its SpecVersion.VersionFlag was removed in 2.0, which ogcapi-foundation now exports, so instantiating the default validator failed with NoClassDefFoundError. Since the server is built lazily on the first request, this surfaced as an HTTP 500 on every call rather than at startup.

Pass our own validator, implemented against the 2.0 API and following SchemaValidatorImpl, so the SDK never loads the default one and no second json-schema-validator version enters the build.

### DIFF
--- a/ogcapi-custom/ogcapi-mcp/src/main/java/de/ii/ogcapi/mcp/app/McpJsonSchemaValidator.java
+++ b/ogcapi-custom/ogcapi-mcp/src/main/java/de/ii/ogcapi/mcp/app/McpJsonSchemaValidator.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2022 interactive instruments GmbH
+ *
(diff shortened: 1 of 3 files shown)


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1770.surge.sh

PR #1772, opened 2026-09-07
CRUD: return a representation of the feature in the response

### LINKED ISSUE #1771: CRUD: support req class "Return Resource Representation in Response"
The requirements class has just been added to the OGC draft.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1771.

A POST, PUT or PATCH request with a `Prefer` header with the value "return=representation" is answered with a representation of the created or changed feature in the response body: 201 with the URI of the feature in a `Location` header for a new feature, 200 for a changed one. Without the preference, or with "return=minimal", the response has no body as before, and the response reports the preference that was applied in a `Preference-Applied` header.

The representation is the returnable representation that a GET request for the feature returns, so it includes the values that the server has assigned, in the coordinate reference system of the `Content-Crs` header of the request. Its media type is negotiated with the `Accept` header among the feature encodings of the collection; a request without an `Accept` header, or one that accepts any media type, is answered in the media type of the request body. The encoding is negotiated down to the subtype only, so `Accept: text/csv` is answered with 406 ("Not Acceptable") rather than with an encoding that merely shares the type. The negotiation happens in the resource method: a request that does not ask for a representation has no response body and is never rejected because of an `Accept` header.

Support is opt-in with the new option `returnRepresentation`, which also drives the declaration of the conformance class "Return Resource Representation in Response" of OGC API - Features - Part 4 (draft).


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1772.surge.sh

PR #1775, opened 2026-09-12
search: composite-key result sets (inResultSetByKey)

### LINKED ISSUE #1774: Search: join a result set on a composite key
Problem: 

A result set declared by a query of a query expression can only be joined on a single value — the ids of the selected features, or the values of one of their properties (`values`). Consumers reference it with `inResultSet`. Feature types whose natural key has several parts cannot be joined that way. 

Proposal:

A second predicate that joins on the whole key at once. The declaring query names the parts of its key, and the consuming query maps the same names to its own properties:

```json
{ "collections": ["parcel"],
  "resultSets": { "s_municipality": { "key": {
      "state": "adm_state", "region": "adm_region",
      "district": "adm_district", "municipality": "adm_municipality" } } } }
```

```json
{ "collections": ["municipality"],
  "filter": { "op": "inResultSetByKey", "args": [
      { "state": { "property": "state_id" }, "region": { "property": "region_id" },
        "district": { "property": "district_id" }, "municipality": { "property": "municipality_id" } },
      "s_municipality" ] } }
```

A result set is either a set of values (`values`, consumed with `inResultSet`) or a set of keys (`key`, consumed with `inResultSetByKey`) — never both, and consuming one as the other is rejected.

Comments on the proposal:

- Parts are matched by name, not by position: A positional key would be shorter, but the parts of a key are typically all of the same type, so transposing two of them validates, runs, and returns the wrong features with
nothing to show for it. With names, a part that one side does not declare is reported before the query is encoded ("defines [district, state], the predicate supplies [municipality, state]"), and the order in the request is irrelevant.

- Key parts should be restricted to columns of the main table, on both sides: A property in a child or junction table encodes as a semi-join subquery, which cannot take part in a row-wise comparison. Both sides should check this against the mapping and name the offending part rather than producing invalid SQL.

- No CQL2-Text encoding: The predicate is only usable inside a query expression, which is JSON, and a named key has no natural text spelling. It should be declared query-expression-only alongside `inResultSet`.

`inResultSet` should be unchanged, so existing query expressions are unaffected.

cportele:
### Pull request checklist

-   [x] Updated documentation

Closes #1774

Depends on https://github.com/ldproxy/xtraplatform-spatial/pull/631

### DIFF
--- a/gradle/layers.versions.toml
+++ b/gradle/layers.versions.toml
@@ -1,5 +1,5 @@
 [versions]
 xtraplatform-core    = '7.1.0-SNAPSHOT'
 xtraplatform-native  = '2.7.0-SNAPSHOT'
-xtraplatform-spatial = '8.1.0-SNAPSHOT'
+xtraplatform-spatial = '8.1.0-inresultsetbykey-SNAPSHOT'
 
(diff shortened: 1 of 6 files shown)


azahnen:
Surge PR preview deployment succeeded. View it at https://ldproxy-ldproxy-pr-1775.surge.sh

azahnen (approved):

